PR #1, opened 2018-03-24
Course scraper files

danielleung27:
course_scraper.py: Modified recal course scraper outputs json of all courses. 
course_scraper_output.json: This is the output of the scraper (scraper takes a while to finish running)


churichard:
It doesn't look like Recal is using this scraper anymore - it gives data from Spring 2015 (which is the course data that the JSON has). Take a look at [this directory](https://github.com/recalapp/recal/tree/master/course_selection) and look at the scrape_*.py files, which is what I think they're using now to scrape the course data.

churichard:
P.S. Also try to have a better commit name, and put your files inside a new directory `courses` instead of in the main directory.

PR #6, opened 2018-03-27
Add dept reqs

bnehoran:
Added json requirement files for AB, BSE, WWS, and COS (both AB and BSE), as well as additional helping files.


scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
+{
+  "type": "degree",
+  "name": "A.B.",
+  "note": "",
+  "courses_needed": 31,
+  "url": "https://ua.princeton.edu/contents/program-study-degree-bachelor-arts",
```
churichard: Please put all the URLs where you gathered information from - this applies to all JSON files

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
...
+  "Requirements": [
+    {
+      "name": "Degree Progress",
+      "max_counted": 1,
+      "min_needed": "",
```
churichard: The better style to represent values that don't exist in JSON is to write `null` (per [the official JSON website](http://www.json.org/)). Please change all the empty strings to `null` instead, both in this file and in all the other files.

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
...
+          "min_needed": "",
+          "description": "",
+          "explanation": "",
+          "completed_by_semester": 8,
+          "NumCourses": 31
+        }
+      ]
+    },
+    {
+      "name": "Writing Seminar",
```
churichard: Technically this needs to be done in freshman year, so you should put a `"completed_by_semester": 2`. Same with the BSE requirements

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
...
+  "courses_needed": 31,
+  "url": "https://ua.princeton.edu/contents/program-study-degree-bachelor-arts",
+  "Requirements": [
+    {
+      "name": "Degree Progress",
+      "max_counted": 1,
+      "min_needed": "",
+      "description": "",
+      "explanation": "With the exception of students who receive one or two terms of advanced standing, all A.B. students must successfully complete a minimum of 31 courses and two years of departmental independent work in eight terms of study. An extra term or year of study is granted by the Faculty Committee on Examinations and Standing to a student making normal progress toward the degree only under extraordinary circumstances.",
+      "ReqList": [
```
churichard: Personally I think all json fields should be in snake_case (even nodes in the graph), so that there's no ambiguity to the programmer as to what kind of casing to use for each field. If you disagree, we can discuss further

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
...
+    },
+    {
+      "name": "Writing Seminar",
+      "max_counted": 1,
+      "min_needed": 1,
+      "description": "Writing Seminar — one course",
+      "explanation": "All Princeton undergraduates take a Writing Seminar in the Fall or Spring of the freshman year.\n\nThe Writing Seminars give Princeton freshmen an early opportunity to belong to a lively academic community in which members investigate a shared topic and discuss their writing together, with the aim of clarifying and deepening their thinking. The seminars are interdisciplinary in nature to emphasize transferable reading, writing and research skills. You’ll learn how to frame compelling questions, position your argument within a genuine academic debate, substantiate and organize claims, purposefully integrate a wide variety of sources and revise for greater cogency and clarity.\n\nEnrollment process:\nhttps://writing.princeton.edu/seminars/how-enroll",
+      "pdf_allowed": false,
+      "CourseList": [
+        "WRI 1**"
```
churichard: How do you know this should be WRI 1**?

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
...
+  "name": "A.B.",
+  "note": "",
+  "courses_needed": 31,
+  "url": "https://ua.princeton.edu/contents/program-study-degree-bachelor-arts",
+  "Requirements": [
+    {
+      "name": "Degree Progress",
+      "max_counted": 1,
+      "min_needed": "",
+      "description": "",
```
churichard: Like the other requirements, this should also have a description

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
...
+      ]
+    },
+    {
+      "name": "Foreign Language",
+      "max_counted": 1,
+      "min_needed": 1,
+      "description": "Foreign Language — one to four terms to complete, depending on the language students study and the level at which they start",
+      "explanation": "When you become proficient in a foreign language, you acquire more than a communication skill; you become literate in another culture and gain another perspective on the world. All candidates for the A.B. degree at Princeton must demonstrate proficiency in a foreign language before graduation. Many undergraduates satisfy the foreign language requirement by demonstrating proficiency when they enter the University. In order to fulfill the language requirement through coursework, we expect successful completion of courses normally numbered through 107/108. ",
+      "double_counting_allowed": false,
+      "Req": "LANG"
```
churichard: Is the foreign language requirement (as well as the other distribution requirements) going to be done through code? Also, why is this `Req` and not `CourseList`?

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
+{
+  "type": "degree",
+  "name": "A.B.",
+  "note": "",
+  "courses_needed": 31,
+  "url": "https://ua.princeton.edu/contents/program-study-degree-bachelor-arts",
+  "Requirements": [
```
churichard: So does `Requirements` just mean that everything underneath it has to be satisfied, i.e. there is an implicit `min_needed` attribute in the root that is equivalent to the length of `Requirements`? Do you think it might be better to specify `min_needed`?

Also, what's the rationale behind having this be named `Requirements` and having the other `ReqList`s be named `ReqList`?

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
...
+        "WRI 1**"
+      ]
+    },
+    {
+      "name": "Foreign Language",
+      "max_counted": 1,
+      "min_needed": 1,
+      "description": "Foreign Language — one to four terms to complete, depending on the language students study and the level at which they start",
+      "explanation": "When you become proficient in a foreign language, you acquire more than a communication skill; you become literate in another culture and gain another perspective on the world. All candidates for the A.B. degree at Princeton must demonstrate proficiency in a foreign language before graduation. Many undergraduates satisfy the foreign language requirement by demonstrating proficiency when they enter the University. In order to fulfill the language requirement through coursework, we expect successful completion of courses normally numbered through 107/108. ",
+      "double_counting_allowed": false,
```
churichard: What does this `double_counting_allowed` field mean?

scripts/deptreqs/BSE_2018.json
```diff
@@ -0,0 +1,371 @@
...
+    },
+    {
+      "name": "Computer Science",
+      "max_counted": 1,
+      "min_needed": 1,
+      "description": "Computer proficiency is a requirement for the B.S.E. degree",
+      "explanation": "Computer proficiency is a requirement for the B.S.E. degree fulfilled by taking COS 126 General Computer Science. This requirement must be satisfied before the beginning of the junior year. This requirement may not be satisfied by a course taken under the pass/D/fail option. A course taken at another school may not be used to satisfy this requirement.",
+      "pdf_allowed": false,
+      "completed_by_semester": 4,
+      "CourseList": [
```
churichard: Remove this `CourseList` - doesn't seem like it should be here (plus it produces a syntax error because there's no comma afterwards)

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
+{
+  "type": "degree",
+  "name": "A.B.",
+  "note": "",
+  "courses_needed": 31,
```
churichard: Probably not needed - it's already encoded elsewhere. You can remove it from the BSE requirements too

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
+{
+  "type": "degree",
+  "name": "A.B.",
+  "note": "",
```
churichard: Do you use this `note` field at all? If not, I think it should be removed (both here and in other files) - it can always be added later if necessary

scripts/deptreqs/BSE_2018.json
```diff
@@ -0,0 +1,371 @@
...
+      "explanation": "B.S.E. students enroll in four courses for the first term of the freshman year and in four or five courses in each succeeding term, following a sequence appropriate to their individual programs. The school requirement for the B.S.E. degree is at least 36 courses in the four years of study. Sophomores, juniors, and seniors must complete at least four courses each term, with a minimum of 17 courses by the start of junior year and 26 courses by the start of senior year.",
+      "ReqList": [
+        {
+          "name": "",
+          "max_counted": 1,
+          "min_needed": "",
+          "description": "",
+          "explanation": "",
+          "completed_by_semester": 1,
+          "NumCourses": 4
```
churichard: It seems to me like four courses is a general minimum course load for each semester, not just for the first semester. You should either encode this minimum for all semesters (as well as encode the minimum course load for each semester for the AB json), or delete it here in order to be consistent with how AB is currently formatted.

scripts/deptreqs/BSE_2018.json
```diff
@@ -0,0 +1,371 @@
...
+  "name": "B.S.E.",
+  "note": "",
+  "courses_needed": 36,
+  "url": "https://ua.princeton.edu/contents/program-study-degree-bachelor-science-engineering",
+  "Requirements": [
+    {
+      "name": "Degree Progress",
+      "max_counted": 1,
+      "min_needed": 4,
+      "description": "",
```
churichard: Like the other requirements, this should have a description

scripts/deptreqs/BSE_2018.json
```diff
@@ -0,0 +1,371 @@
...
+        {
+          "name": "Third Term Math",
+          "max_counted": 1,
+          "min_needed": 1,
+          "description": "",
+          "explanation": "201 Multivariable Calculus, 203 Advanced Multivariable Calculus, 218 Analysis in Several Variables, or EGR 192 Integrated Introduction to Engineering, Mathematics, Physics",
+          "CourseList": [
+            "MAT 201",
+            "MAT 203",
+            "MAT 216",
```
churichard: The requirements don't list MAT 216 as one of the possible courses to fulfill this

scripts/deptreqs/BSE_2018.json
```diff
@@ -0,0 +1,371 @@
...
+            "PHY 105",
+            "EGR 191"
+          ]
+        },
+        {
+          "name": "Electromagnetism",
+          "max_counted": 1,
+          "min_needed": 1,
+          "description": "",
+          "explanation": "104 General Physics or 106 Advanced Physics, or equivalent",
```
churichard: Do we want to have some way to specify equivalent courses more precisely?

We can list the actual equivalent courses in the json, or we can just have some kind of signifier to say that there *are* equivalent courses, and then scrape them from the course offerings.

I think the second way is cleaner, but we'd have to be careful with how we get the equivalent courses. For example, ISC fulfills the chem requirements but is not listed as an equivalent course in CHM 201. However, CHM 201 is mentioned in the ISC course offerings page as one of the classes that ISC can count for (in the Other Information section, even - not in the Equivalent Courses section!)

If we do opt for the second option, we can have a ` - E` next to the course name to say that an equivalent course of it can be taken, or we can make each course name an object with various attributes instead of just a string

Perhaps the best solution would be to do the second option, but also put in courses like ISC into the json, since they're unique courses. What do you think?

scripts/deptreqs/BSE_2018.json
```diff
@@ -0,0 +1,371 @@
...
+      "ReqList": [
+        {
+          "name": "",
+          "max_counted": 1,
+          "min_needed": "",
+          "description": "",
+          "explanation": "",
+          "CourseList": [
+            "CHM 201",
+            "CHM 201A",
```
churichard: CHM 201A and CHM 203 aren't in the requirements list I'm looking at

scripts/deptreqs/BSE_2018.json
```diff
@@ -0,0 +1,371 @@
...
+        },
+        {
+          "name": "Electromagnetism",
+          "max_counted": 1,
+          "min_needed": 1,
+          "description": "",
+          "explanation": "104 General Physics or 106 Advanced Physics, or equivalent",
+          "CourseList": [
+            "PHY 104",
+            "PHY 106"
```
churichard: I think ISC can fulfill the requirements for Physics as well

scripts/deptreqs/BSE_2018.json
```diff
@@ -0,0 +1,371 @@
...
+      "ReqList": [
+        {
+          "name": "",
+          "max_counted": 1,
+          "min_needed": "",
+          "description": "",
+          "explanation": "",
+          "CourseList": [
+            "COS 126",
+            "COS 217",
```
churichard: Can COS 217 and COS 226 also fulfill this requirement?

scripts/deptreqs/AB_2018.json
```diff
@@ -0,0 +1,120 @@
...
+      "name": "Foreign Language",
+      "max_counted": 1,
+      "min_needed": 1,
+      "description": "Foreign Language — one to four terms to complete, depending on the language students study and the level at which they start",
+      "explanation": "When you become proficient in a foreign language, you acquire more than a communication skill; you become literate in another culture and gain another perspective on the world. All candidates for the A.B. degree at Princeton must demonstrate proficiency in a foreign language before graduation. Many undergraduates satisfy the foreign language requirement by demonstrating proficiency when they enter the University. In order to fulfill the language requirement through coursework, we expect successful completion of courses normally numbered through 107/108. ",
+      "double_counting_allowed": false,
+      "Req": "LANG"
+    },
+    {
+      "name": "Epistemology and Cognition",
```
churichard: Maybe you should make a `Req` for Distribution Requirements and then put all of the distribution requirements inside, in order to parallel how you do it for BSE

churichard (changes requested):
Nice work! So far, I've only reviewed the AB and BSE requirements (which took a long time even by themselves).

What you should do:
- Get rid of the "scripts" folder and just have one folder named "dept_reqs"
- Remove everything except for the AB and BSE jsons from this pull request, and put them in separate pull requests separated out nicely, e.g. by major. You'll have to make separate branches for each pull request.
- Make the changes specified in my comments below

Please also document the JSON format (doesn't need to be in this pull request, you can do it in a different one), and submit smaller pull requests in the future.

scripts/deptreqs/AB_2018.json
bnehoran: Done

scripts/deptreqs/AB_2018.json
bnehoran: Done

scripts/deptreqs/AB_2018.json
bnehoran: done

scripts/deptreqs/AB_2018.json
bnehoran: Done

scripts/deptreqs/AB_2018.json
bnehoran: https://writing.princeton.edu/seminars/fall-seminars

bnehoran: Done

scripts/deptreqs/AB_2018.json
bnehoran: Yes. It is unfeasible to write out all the course combinations that could satisfy it.

scripts/deptreqs/AB_2018.json
bnehoran: Changed to req_list

bnehoran: Yes, the very top req_list has an implicit min_needed = "ALL"

scripts/deptreqs/AB_2018.json
bnehoran: It indicates that courses in this subtree cannot be counted towards more than one branch.

scripts/deptreqs/BSE_2018.json
bnehoran: Done

scripts/deptreqs/AB_2018.json
bnehoran: Done

scripts/deptreqs/AB_2018.json
bnehoran: Done

scripts/deptreqs/BSE_2018.json
bnehoran: Not sure how to encode this "courses per semester" within the existing format...

bnehoran: Also, the AB course-load requirements seem to be a bit more complicated with senior thesis factoring in.

scripts/deptreqs/BSE_2018.json
bnehoran: Not sure what to put there.

bnehoran: Also, I have a feeling that we might want to eliminate the "description" field and just merge it into the other two.

scripts/deptreqs/BSE_2018.json
bnehoran: from Degree Progress Report

scripts/deptreqs/BSE_2018.json
bnehoran: Not sure. The second option seems complicated to implement, as I don't know how feasible it would be to automatically detect equivalent courses (as opposed to prerequisite courses etc.
I think we need to discuss this more in detail.

scripts/deptreqs/BSE_2018.json
bnehoran: from Degree Progress Report

scripts/deptreqs/BSE_2018.json
bnehoran: Done

scripts/deptreqs/BSE_2018.json
bnehoran: Apparently, according to the Degree Progress report

scripts/deptreqs/AB_2018.json
bnehoran: I think it's based on the way that they're listed on the respective requirements page. For BSE, for example, it's listed as a Humanities and Social Sciences requirement, whereas AB doesn't do that.
But maybe...

scripts/deptreqs/BSE_2018.json
churichard: Okay, we'll keep it as is for now.

scripts/deptreqs/AB_2018.json
churichard: Let's stick with how it is for now then

churichard (approved):

PR #9, opened 2018-04-09
Added course scraper files

danielleung27:
Added models for courses, course scraper python files, command to run scraper, global variables to base.py settings file, and cache functions for course data in production.py


config/settings/base.py
```diff
@@ -117,3 +117,19 @@
...
+# For example, the Fall 2016 term is 1172 - the ending year is 2017.
+# Similarly, the Spring 2017 term is 1174 - the ending year is 2017.
+# Note: generally, limit this to three terms. This defines the terms that are
+#   scraped and the default terms displayed on schedules.
+ACTIVE_TERMS = [1182, 1184]
+
+# TODO: is this still used? if not, we can remove it
+# Use helper method nice.models.get_cur_semester() to get current Semester
+# object.
+CURR_TERM = 1184
```
churichard: I think these lines can be removed - doesn't seem like we need to have a notion of what the `CURR_TERM` is (and even if we do, we can simply find the max of the `ACTIVE_TERMS`

config/settings/development.py
```diff
@@ -1,5 +1,4 @@
 from .base import *
 
-
```
churichard: Don't make superfluous changes

config/settings/production.py
```diff
@@ -1,13 +1,63 @@
 from .base import *
 import os
+from os import environ
 
 
-DEBUG = False
+#DEBUG = False
+DEBUG = environ.get('DJANGO_DEBUG', False)
```
churichard: I don't think you need to change this

config/settings/production.py
```diff
@@ -1,13 +1,63 @@
...
 ADMIN_ENABLED = DEBUG
 
-ALLOWED_HOSTS = os.getenv('ALLOWED_HOSTS', '*').split(',')
+#ALLOWED_HOSTS = os.getenv('ALLOWED_HOSTS', '*').split(',')
+ALLOWED_HOSTS = environ.get('ALLOWED_HOSTS', '*').split(',')
```
churichard: I don't think you need to change this

config/settings/production.py
```diff
@@ -1,13 +1,63 @@
...
 
 
+def get_cache():
```
churichard: Is this caching stuff necessary for the code to run? I don't think we have memcache set up anyway in Heroku

tigerpath/management/commands/access_cache.py
```diff
@@ -0,0 +1,11 @@
+from django.conf import settings
```
churichard: Is this caching stuff necessary for the code to run?

tigerpath/models.py
```diff
@@ -1,3 +1,230 @@
...
+    start_date = models.DateField()
+    end_date = models.DateField()
+    """ term_code = 1xxy, where xx is the year in which the school year ends,
+    and y is the semester code. y = 2 for the fall term, y = 4 for the spring
+    Example:
+    1144 = 1314Spring
+    1132 = 1213Fall
+    """
+    term_code = models.CharField(
+        max_length=4, default=settings.CURR_TERM, db_index=True, unique=True)
```
churichard: You may need to get rid of the default argument if you remove `CURR_TERM` from settings

tigerpath/models.py
```diff
@@ -1,3 +1,230 @@
...
+        else:
+            sem = 'Spring'
+        return str(start_year) + "-" + str(end_year) + " " + sem
+
+
+class Professor(models.Model):
+    name = models.CharField(max_length=100)
+
+
+class Color_Palette(models.Model):
```
churichard: Is this necessary?

tigerpath/models.py
```diff
@@ -1,3 +1,230 @@
...
+    end_time = models.CharField(max_length=20)
+    days = models.CharField(max_length=10)
+    location = models.CharField(max_length=50)
+
+    def __unicode__(self):
+        return unicode(self.section) + ' - ' + self.location
+
+
+class Course_Listing(models.Model):
+    # TODO: this line causes admin site to fail, commenting out related_name
```
churichard: Does the admin site work? If so, maybe you can get rid of this TODO comment (doesn't seem like we should be pushing in TODO comments if we can help it)

tigerpath/models.py
```diff
@@ -1,3 +1,230 @@
...
+    is_primary = models.BooleanField(default=False)
+
+    def __unicode__(self):
+        return self.dept + ' ' + self.number
+
+    class Meta:
+        ordering = ['dept', 'number']
+
+'''
+class Schedule(models.Model):
```
churichard: Looks like all of this stuff that's commented is probably unnecessary, you can probably remove it

churichard (changes requested):
Nice work - there are a few comments that I made in the code itself, and a general comment below.

How did you create all_courses.json? I ask because the first few lines in the json are 
```
course selection: courses scraped successfully
course selection: cache regenerated for term 1182
course selection: cache regenerated for term 1184
course selection: courseapi cache cleared.
```
and then the actual json starts. Is this json supposed to be used somewhere? Does it work as is? And does it contain all the courses from the terms that are in `ACTIVE_TERMS` (i.e. Fall 2017 and Spring 2018)?

config/settings/base.py
danielleung27: I can remove the comments. I think we should keep the CURR_TERM variable. If we're in the fall semester then we would want the smaller term so I don't think we could just take the max everytime. Also it might be better for clarity since it's not really obvious how the term is created without going back to base.py and reading the comment

config/settings/production.py
danielleung27: I changed up the structure in production.py to use os.environ instead so it would be easier to copy over the stuff from recal for the caching. The scraper works without the caching but I wasn't sure if we needed the caching or not so I implemented the caching anyways. Should I just remove caching?  I can change production.py back to the way it was if caching is not necessary.

danielleung27:
The all_courses.json is just the raw output of the scraper in case any of us wants to look at it without going through the database where each course is fragmented into different models. It contains all the course data from the active terms.

config/settings/base.py
churichard: Maybe I'm misunderstanding, but I think the max would work. For example, Fall 2016 is 1172 and Spring 2017 is 1174. Once it turns Fall 2017, the code would actually be 1182 (since it uses the last two digits of the ending year, and Fall 2017 is part of 2017-2018). Thus, it seems like the term codes are monotonically increasing.

However, you're probably right that having an explicit `CURR_TERM` would be clearer to read. I'm just not sure if there is a scenario in which our app would need to know what the current term is. What do you think?

config/settings/production.py
churichard: I think for now we should leave out the caching (since I don't know how it works anyway). If it's too slow we can then think about implementing it

churichard:
Okay cool. You might want to remove the first four lines, since they're not technically part of the json. Also, you should rename it to clarify which terms the courses are from.

config/settings/base.py
danielleung27: Oh I see using the max would work. I was thinking we would add both fall and spring semesters at once but I realize we can't scrape future semesters. I grepped all the files in recal and it seems like CURR_TERM is only used in that one instance you marked below in models.py for the Semester class. We could probably get rid of the global variable.

tigerpath/views.py
```diff
@@ -33,3 +38,64 @@ def landing(request):
 # about page
 def about(request):
     return render(request, 'tigerpath/about.html', None)
+
```
churichard, line 41: Can you put two blank lines in between each function, for consistency?

churichard (changes requested):
There's a few extra changes that need to be made:

- Move all the `scraper_*.py` files to a folder called `scraper`. You'll also need to change the import of `scrape_all` in `tigerpath_get_courses.py`
- Add the following lines to `requirements.txt`:
    - `beautifulsoup4==4.6.0`
    - `bs4==0.0.1`
    - `lxml==4.2.1`
- In `scrape_parse.py`, change `BeautifulSoup(seed_page)` to `BeautifulSoup(seed_page, "lxml")` in order to suppress warnings

tigerpath/scraper/scrape_import.py
```diff
@@ -26,10 +26,10 @@ def __str__(self):
 
 
 def scrape_import_course(course, counter=ScrapeCounter()):
-    from .models import Course
+    from ..models import Course
```
churichard: Maybe tigerpath.models would be clearer for all of these imports

churichard (dismissed):
Great job! After you change this one thing, you can then squash and merge

churichard (approved):

PR #10, opened 2018-04-13
Add requirements JSONs for WWS, COS-AB, and COS-BSE

bnehoran:


majors-and-certificates/majors/COS-AB_2018.json
```diff
@@ -0,0 +1,240 @@
...
+            "COS 325/MUS: 315 Transforming Reality by Computer",
+            "COS 326: Functional Programming",
+            "COS 401/TRA 301: Introduction to Machine Translation ",
+            "COS 402: Machine Learning and  Artificial Intelligence",
+            "COS 424: Fundamentals of Machine Learning",
+            "COS 426: Computer Graphics",
+            "COS 429: Computer Vision",
+            "COS 432: Information Security",
+            "COS 435: Information Retrieval, Discovery, and Delivery",
+            "COS 436: Human-Computer Interface Technology ",
```
churichard: There's an extra space at the end here

majors-and-certificates/majors/COS-AB_2018.json
```diff
@@ -0,0 +1,240 @@
...
+            "MOL 437/NEU 437",
+            "NEU 330",
+            "ECO 326"
+          ]
+        }
+      ]
+    },
+    {
+      "name": "Independent Work",
+      "max_counted": null,
```
churichard, line 233: For independent work, should we still check for it and the user can manually tick off that they've satisfied it, or should we not check for it at all?

Also, are AB COS people not allowed to partake in an independent work seminar?

majors-and-certificates/majors/COS-BSE_2018.json
```diff
@@ -0,0 +1,246 @@
...
+    {
+      "name": "Independent Work",
+      "max_counted": 1,
+      "min_needed": 1,
+      "description": "Independent Work",
+      "explanation": "All B.S.E. concentrators engage in independent work supervised by a member of the department, often associated with a research project. It may require a significant programming effort, a theoretical study involving the design and analysis of algorithms, or an applications problem in some other field. The results of these efforts must be presented in a written report and poster session. B.S.E. students must elect one semester of independent work by enrolling in 397, 398, 497, or 498. One additional semester of independent work may be counted as one of the departmental courses.",
+      "double_counting_allowed": false,
+      "pdfs_allowed": false,
+      "course_list": [
+        "COS 397",
```
churichard, line 239: If you do independent work as COS BSE, do you have to enroll in one of these courses even if you're not doing a seminar?

majors-and-certificates/majors/WWS_2018.json
```diff
@@ -0,0 +1,1292 @@
...
+              "explanation": "Human Welfare and Social Policy\nEducation\nGender, Sexuality, and Families\nHealth and Well Being\nHousing and Urban Issues\nHuman Rights\nImmigration\nPoverty, Inequality, Income, and Social Mobility\nRace, Ethnicity, and Discrimination",
+              "req_list": [
+                {
+                  "name": "Education",
+                  "max_counted": null,
+                  "min_needed": null,
+                  "description": "Education",
+                  "explanation": "Education",
+                  "course_list": [
+                    "AMS 311: Education and Inequality",
```
churichard, line 198: Why are you putting the elective courses here if you also have them in a separate file?

On that note, I just thought of something - what do you think of separating the courses that satisfy requirements from the actual json itself? If we separate them, the json would encode the structure of the requirements (which may change, but probably relatively infrequently), and there would be a separate file to store the courses (which would probably change more frequently). This would also mean that if someone were to update courses, they wouldn't need to dig through the structure of the json in order to do so. We already do this to a certain extent, like for LANG and distribution requirements - doing it for everything might increase maintainability.

Let me know what you think

majors-and-certificates/majors/WWS_2018.json
```diff
@@ -0,0 +1,1292 @@
...
+                {
+                  "name": "Middle East",
+                  "max_counted": null,
+                  "min_needed": null,
+                  "description": "Middle East",
+                  "explanation": "Middle East",
+                  "course_list": [
+                    "HIS 267: The Modern Middle East",
+                    "HIS 419: Topics in History of Modern Syria: Ba’athist Syria – Ideology, Literature & Film",
+
```
churichard: Earmarking this with the same issue raised above:

For senior thesis, should we still check for it and the user can manually tick off that they've satisfied it, or should we not check for it at all?

churichard: Remove note

churichard: Remove note

churichard: Remove note

On that note, how are you checking these foreign language requirements in the req checker (i.e. how are you enforcing that LANG has to be different between these two inner branches (that check LANG 102, LANG 103, etc.) but that the LANG in these inner branches _can_ be the same as the LANG that checks for LANG 2** and LANG 3**)?

churichard: Earmarking this with the same issue raised before:

For field experience, should we still check for it and the user can manually tick off that they've satisfied it, or should we not check for it at all?

churichard (changes requested):
Nice work! I raised a couple of issues below.

majors-and-certificates/majors/COS-BSE_2018.json
bnehoran: yes

majors-and-certificates/majors/COS-AB_2018.json
bnehoran: Done

majors-and-certificates/majors/COS-AB_2018.json
bnehoran: No. Only BSE students sign up for COS 397/398/497/498.

bnehoran: It will appear, and will be unsatisfiable, but also not required.

majors-and-certificates/majors/WWS_2018.json
bnehoran: I think it might be useful to add that as an option.
I think I'll just make a note, and we can add it in later, since I don't think it's high priority.

majors-and-certificates/majors/WWS_2018.json
bnehoran: It's in place of a comment.

bnehoran: Not sure. Need to think about it.

bnehoran: Same response.

churichard (approved):

PR #11, opened 2018-04-15
Add course search

danielleung27:
/: webpack-stats files necessary for react. 

/tigerpath: Added urls to urls.py for search engine page (will integrate with front end later) and get request. Added get request function to views.py (get_courses) and implemented filter functions which creates a filtered list of courses to send to react end. Resources.py is currently not used but it is used for tastypie api just in case we need it later.

/tigerpath/templates/tigerpath: added search.html to display course searcher, will integrate with front end later 

/New folder: all files here are autogenerated bundles from react (I am not sure which file alters what this folder is named but currently it is automatically named New folder)

/frontend: all files here are for react. 

/frontend/src: These files are the more important ones. Search.js is the react code for the searching and Courses.js is a child to Search which displays the course cards. App.css and index.css are there for reference and will be modified later to style the react components. I'm not sure what registerServiceWorker.js is, it was provided when I set up react.


config/settings/base.py
```diff
@@ -17,8 +18,16 @@
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'django_cas_ng',
+    'webpack_loader',
 ]
 
+WEBPACK_LOADER = {
```
churichard: Should probably be in development.py instead

config/settings/production.py
```diff
@@ -7,6 +7,16 @@
 
 ALLOWED_HOSTS = os.getenv('ALLOWED_HOSTS', '*').split(',')
 
+STATICFILES_DIRS = [
```
churichard: Should probably be in base.py

frontend/public/index.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
```
churichard, line 1: Is the stuff in frontend/public needed? I think we're probably serving stuff from the Django server, right?

frontend/src/Courses.js
```diff
@@ -0,0 +1,25 @@
+import React, { Component } from 'react';
+import './App.css';
+
+import $ from 'jquery'
+import jQuery from 'jquery'
+
+class Courses extends Component {
+  constructor()
+  {
```
churichard: Let's use the style of putting the brace on the same line as the function name (change this in the other js files you modified as well):
```
function() {
    # code here
}
```

frontend/README.md
```diff
@@ -0,0 +1,2444 @@
+This project was bootstrapped with [Create React App](https://github.com/facebookincubator/create-react-app).
```
churichard: Delete this file

New folder/New folder/asset-manifest.json
```diff
@@ -0,0 +1,6 @@
+{
```
churichard: Delete all this New folder stuff

frontend/.gitignore
```diff
@@ -0,0 +1,21 @@
+# See https://help.github.com/ignore-files/ for more about ignoring files.
```
churichard, line 1: Move this into the root level .gitignore

frontend/src/Courses.js
```diff
@@ -0,0 +1,25 @@
...
+    super();
+  }
+  render() {
+    console.info(this.props.course_list)
```
churichard: Don't print this out

tigerpath/urls.py
```diff
@@ -9,4 +9,6 @@
     # cas auth
     path('login', views.login, name='cas_ng_login'),
     path('logout', views.logout, name='cas_ng_logout'),
+    path('search', views.search, name='search'),
+    path('get_courses/<search_query>', views.get_courses, name='get_courses'),
```
churichard: This is part of the API right? I think a better URL might be 'api/v1/get_courses/<search_query>' (it makes it clear that it's part of the API and if we wanted, we could add a new URL in the future for v2 without breaking any apps that are using v1)

webpack-stats.dev.json
```diff
@@ -0,0 +1 @@
...tus":"done","publicPath":"http://localhost:3000/","chunks":{"main":[{"name":"static/js/bundle.js","publicPath":"http://localhost:3000/static/js/bundle.js","path":"C:\\Users\\Daniel\\Desktop\\Classwork\\Sophomore\\S2\\Cos 333\\TigerPath\\tigerpath\\frontend\\static\\js\\bundle.js"},{"name":"0.31965e05e4730bcaa338.hot-update.js","publicPath":"http://localhost:3000/0.31965e05e4730bcaa338.hot-update.js","path":"C:\\Users\\Daniel\\Desktop\\Classwork\\Sophomore\\S2\\Cos 333\\TigerPath\\tigerpath\\frontend\\0.31965e05e4730bcaa338.hot-update.js"},{"name":"static/js/bundle.js.map","publicPath":"http://localhost:3000/static/js/bundle.js.map","path":"C:\\Users\\Daniel\\Desktop\\Classwork\\Sophomore\\S2\\Cos 333\\TigerPath\\tigerpath\\frontend\\static\\js\\bundle.js.map"},{"name":"0.31965e05e4730bcaa338.hot-update.js.map","publicPath":"http://localhost:3000/0.31965e05e4730bcaa338.hot-update.js.map","path":"C:\\Users\\Daniel\\Desktop\\Classwork\\Sophomore\\S2\\Cos 333\\TigerPath\\tigerpath\\frontend\\0.31965e05e4730bcaa338.hot-update.js.map"}]},"error":"ModuleError","message":"\n  Line 42:  'Courses' is not defined  react/jsx-no-undef\n\nSearch for the keywords to learn more about each error."}
```
churichard: I don't particularly like the fact that this is tied to a path on your local computer, is there some way to fix this?

tigerpath/views.py
```diff
@@ -40,11 +41,103 @@ def about(request):
...
+            results = filter_dept(query, results)
+
+        # is course number
+        elif(len(query) <= 3 and query.isdigit() or len(query) == 4 and query[:3].isdigit()):
+            results = filter_num(query, results)
+
+        # check if it matches title
+        else:
+            # convert course_listings to courses to filter
+            results = listing_to_course(results)
```
churichard: Does it take a long time to convert between courses and listings? This portion of the code seems a bit inefficient

tigerpath/views.py
```diff
@@ -40,11 +41,103 @@ def about(request):
...
+    split_query = re.split('(\d.*)', search_query)
+    queries = []
+    # split again by spaces
+    for query in split_query:
+        queries = queries + query.split(" ")
+    return queries
+
+
+# filter courses where query is a dept code
+def filter_dept(query, results):
```
churichard: For these filtering functions, I think you can use Python's `filter` and pass in lambdas: http://book.pythontips.com/en/latest/map_filter.html#filter

tigerpath/views.py
```diff
@@ -40,11 +41,103 @@ def about(request):
...
+    return results
+
+
+# preprocesses query for filtering
+def break_query(search_query):
+    # split only by first digit occurrance ex: cee102a -> [cee, 102a]
+    split_query = re.split('(\d.*)', search_query)
+    queries = []
+    # split again by spaces
+    for query in split_query:
```
churichard: I think the way you structure these regular expressions means that empty strings are occasionally added to the list of queries, e.g. when you split by spaces for "CEE " and "102A". I noticed you check for this when you're iterating through the queries but I think it would be cleaner to filter them out in this function (you can try using `filter(None, queries)`)

tigerpath/views.py
```diff
@@ -40,11 +41,103 @@ def about(request):
...
+# converts list of course_lisiting models to course models (through foreign key) and vice versa
+def listing_to_course(results, reverse = False):
+    converted_results = []
+    if(not reverse):
+        for course_listing in results:
+            if(course_listing.course not in converted_results):
+                converted_results.append(course_listing.course)
+    else:
+        for course in results:
+            for course_listing in course.course_listing_set.all():
```
churichard: I think you can do `converted_results.extend(course.course_listing_set.all())` instead of having two for loops

churichard (changes requested):
Good work! Make sure to address the comments and also push your updated searching code.

frontend/public/index.html
danielleung27: For some reason, deleting these files causes an error in {% render_bundle 'main' %} line 12 in search.html

webpack-stats.dev.json
danielleung27: This file is autogenerated when you start the react server

tigerpath/views.py
danielleung27: The first part is in case the user combines the department and number. The for loop to split by spaces is if there are any extra queries. Like if the user did "cee102a eng" it will first split into ['cee', '102a eng'] and then it will go through and split into ['cee', '102a', 'eng']

webpack-stats.dev.json
```diff
@@ -1 +1 @@
...
...pdate.js","publicPath":"http://localhost:3000/0.a21e8aeea634b0bb4fbd.hot-update.js","path":"C:\\Users\\Daniel\\Desktop\\Classwork\\Sophomore\\S2\\Cos 333\\TigerPath\\tigerpath\\frontend\\0.a21e8aeea634b0bb4fbd.hot-update.js"},{"name":"static/js/bundle.js.map","publicPath":"http://localhost:3000/static/js/bundle.js.map","path":"C:\\Users\\Daniel\\Desktop\\Classwork\\Sophomore\\S2\\Cos 333\\TigerPath\\tigerpath\\frontend\\static\\js\\bundle.js.map"},{"name":"0.a21e8aeea634b0bb4fbd.hot-update.js.map","publicPath":"http://localhost:3000/0.a21e8aeea634b0bb4fbd.hot-update.js.map","path":"C:\\Users\\Daniel\\Desktop\\Classwork\\Sophomore\\S2\\Cos 333\\TigerPath\\tigerpath\\frontend\\0.a21e8aeea634b0bb4fbd.hot-update.js.map"}]},"error":"ModuleError","message":"\n  Line 27:  Parsing error: Unexpected token, expected )\n\n  25 |         success: function(data) {\n  26 |           this.setState({data: data});\n> 27 |           if(search_query == this.state.search or search_query == '$')\n     |                                                ^\n  28 |             ReactDOM.render(<Courses course_list={data}/>, document.getElementById('courses'));\n  29 |         }.bind(this)\n  30 |     })\n\n"}
```
churichard: If this is autogenerated, I think you can leave it out, right?

churichard (approved):
How do you run this? Do you need to run a special command for React or can you just run the Django server and it'll work? If there are some special steps involved, can you document them in the readme?

Also, make sure to look at the comment I made. Otherwise, looks good. 👍Remember to squash and merge.

webpack-stats.dev.json
danielleung27: Yeah I'll leave it out. For development, you have to go to the folder "frontend" and run 'npm start'. Then you can run the django server and it'll work. For production you do something where the react thing will run from just starting the django server but I wasn't able to test that out because our production settings weren't working yet.

PR #13, opened 2018-04-20
Integrate frontend with backend, added drag/drop

danielleung27:
Integrated search engine with frontend, added drag drop feature, added api for post request for updating schedule in backend.

Courses.js is not needed anymore but keeping the file just in case. Will probably delete soon.


frontend/src/Search.js
```diff
@@ -1,22 +1,60 @@
...
+      document.getElementById("sem4"), 
+      document.getElementById("sem5"), 
+      document.getElementById("sem6"), 
+      document.getElementById("sem7"), 
+      document.getElementById("sem8")]);
+
+    // tells react to update course schedule when an item is dropped
+    drake.on('drop', function (){
+      let added_courses = document.querySelectorAll(".course");
+      let courses_taken = [[],[],[],[],[],[],[],[]];
```
churichard: Instead of hardcoding the semester here, you can make `courses_taken = []` and then under `added_courses.forEach(function (semester){` you can put `courses_taken.push([])`

frontend/.gitignore
```diff
@@ -0,0 +1,132 @@
...
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+package-lock.json
```
churichard: Remove these three lines (I added them before but they're actually not necessary)

frontend/src/Courses.css
```diff
@@ -0,0 +1,42 @@
...
+  background-color: #ffc0cb;
+  list-style-type: none;
+  color: #DB7093;
+}
+
+li:hover{
+  cursor: pointer;
+}
+
+.gu-mirror {
```
churichard: I think most of these gu classes/properties are from dragula, right? A cleaner way to import them might be to do `import styles from 'dragula/dist/dragula.css';` in Courses.js and Search.js (according to https://github.com/bevacqua/react-dragula/issues/27) and then only specify the specific properties you want to add/override in Courses.css.

frontend/src/Search.js
```diff
@@ -1,22 +1,60 @@
 import React, { Component } from 'react';
 import ReactDOM from 'react-dom';
-import Courses from './Courses';
+import './Courses.css';
 import $ from 'jquery'
 import jQuery from 'jquery'
```
churichard, line 5: Don't think you need to import `jQuery` if you're already importing `$`

tigerpath/templates/tigerpath/index.html
```diff
@@ -47,58 +49,10 @@
...
-              <td class="course">  </td>
-            </tr>
-          </tbody>
-          <tbody>
-            <tr>
-              <td class="course">  </td>
-              <td class="course">  </td>
-              <td class="course">  </td>
-              <td class="course">  </td>
+              <td id="sem1" class="course">  </td>
```
churichard: I don't think it makes sense for the css class of a semester to be `course`. Change this to `semester` and change the class in the css file to `semester` as well

If you change the draggable logic, then you can probably get rid of the ids as well

frontend/src/Search.js
```diff
@@ -1,22 +1,60 @@
...
 class Search extends Component {
   constructor() {
     super();
     this.state = {
       search: '',
       data: []
     };
+
+    // select containers to make items draggable
+    let drake = dragula([document.getElementById("display_courses"),
```
churichard: If you wanted this to be cleaner, you can do something like this with jquery:
```javascript
let draggableItems = $(".semester").get()
draggableItems.append($("#display_courses")[0])
let drake = dragula(draggableItems)
```
This assumes that you changed the css class to `semester`.

frontend/src/Search.js
```diff
@@ -1,22 +1,60 @@
...
+      let i = 0;
+      added_courses.forEach(function (semester){
+        semester.childNodes.forEach(function(course){
+          if(typeof course.innerHTML != 'undefined')
+            courses_taken[i].push(course.innerHTML);
+        })
+        i++
+      });
+      courses_taken = JSON.stringify(courses_taken);
+      console.info(courses_taken);
```
churichard: Remove this

frontend/src/Courses.js
```diff
@@ -1,18 +1,44 @@
...
+var dragula = require('react-dragula')
 
 class Courses extends Component {
   constructor() {
     super();
   }
+  componentDidMount() {
+//    let draggables = [document.getElementById("course_list")];
+//    draggables.concat(Array.from(document.querySelectorAll(".course"));
+    let drake = dragula([document.getElementById("course_list"),
```
churichard: Can change this logic to the modified logic in Search.js

tigerpath/templates/tigerpath/index.html
```diff
@@ -192,5 +98,6 @@
         });
       }
     </script>
```
churichard, line 98: I can't quite comment on the correct line, but you're not using this draggable bundle, right? I think you can get rid of the line `<!-- draggable bundle -->` and the one below it

tigerpath/views.py
```diff
@@ -90,6 +85,13 @@ def filter_courses(search_query):
     return results
 
 
+# updates users schedules with added courses
+@csrf_exempt
```
churichard: Does this actually need to be `csrf_exempt`?

churichard (changes requested):
Looks neat! I added a few comments.

There are some things that need to be addressed in order for this to be fully usable (but doesn't have to be in this pull request):

- Need to add a scrollbar for the course searching
- Need to protect the api endpoints somehow (right now anyone can access them, which is a problem) (may need to use tastypie)
    - Also need to remove CSRF exempt (may need to follow these steps: https://docs.djangoproject.com/en/2.0/ref/csrf/#acquiring-the-token-if-csrf-use-sessions-is-false)
- Allow adding the same course multiple times (which is the current behavior), but show a tooltip saying that the course is a duplicate if it already exists in the schedule. There are some courses (like CWR 202) which you can take multiple times.
- Add "X" button to remove courses from a semester
- Add course name to each card in the search results
- Add a django-admin command to combine semester data from past four years into one comprehensive course list (with typical times that the course is offered, etc.)
- Prevent users from moving a course in their schedule back to the course search
- Prevent users from reordering courses in the course search

tigerpath/urls.py
```diff
@@ -9,6 +9,6 @@
     # cas auth
     path('login', views.login, name='cas_ng_login'),
     path('logout', views.logout, name='cas_ng_logout'),
-    path('search', views.search, name='search'),
     path('api/v1/get_courses/<search_query>', views.get_courses, name='get_courses'),
```
churichard: Also, add a comment above this line: `# api`

frontend/src/Search.js
```diff
@@ -15,40 +15,46 @@ class Search extends Component {
...
       let i = 0;
       added_courses.forEach(function (semester){
+        courses_taken.push([]);
         semester.childNodes.forEach(function(course){
-          if(typeof course.innerHTML != 'undefined')
+          if(typeof course.innerHTML != 'undefined'){
             courses_taken[i].push(course.innerHTML);
+            // adds onclick listener to remove courses on schedule
+            // .off to prevent multiple bindings of click event
+            $("#" + course["id"]).off("click").click(function(){
```
churichard: Courses should not be removed by simply clicking on them anywhere - this makes it way too easy to accidentally remove courses. You should have an explicit `X` button on the right side that will remove a course when clicked

Also, doesn't make much sense to have this onclick listener in this function. It should be separated out and only added once (that way you won't need to have the `.off("click")` call)

frontend/src/Search.js
```diff
@@ -15,40 +15,46 @@ class Search extends Component {
...
+                $("#display_courses").append($(this)[0].outerHTML);
```
churichard: Don't add the course back to the course listings. Doesn't make much sense to do that

frontend/src/Search.js
```diff
@@ -15,40 +15,46 @@ class Search extends Component {
...
+    let draggableItems = $(".semester").get()
+    draggableItems.push($("#display_courses")[0])
+    let drake = dragula(draggableItems)
 
-    // tells react to update course schedule when an item is dropped
-    drake.on('drop', function (){
-      let added_courses = document.querySelectorAll(".course");
-      let courses_taken = [[],[],[],[],[],[],[],[]];
+    // gets current enrolled courses and sends post request
+    let get_courses = function(){
```
churichard: JavaScript uses camelCase as the convention for variable names and function names.

Also, this is not the most descriptive function name. A better one might be `updateSchedule` (since that's what you're doing). The get courses logic might also be able to be separated out into another function

churichard (changes requested):

frontend/src/Search.js
danielleung27: I put the onclick listener in the updateSchedule function so that only the courses in the schedule can be deleted (and not the ones in the search). Once I add the X's on the courses I think I can make this cleaner by selecting all X elements.

frontend/src/Search.js
danielleung27: I was thinking if a user searches Cos and adds a couple classes and then removes some, the removed Cos courses should go back to the search list. It might be weird if they have Ele searched and then they remove a cos class and it goes to the search, but I don't think it is as annoying as if they were to search Cos, add 3 classes, remove one, and then have to re-search Cos if they want to re-add the one they removed.

churichard: In that case, I think it might be best to have it so that every time the user removes a course, it checks to see if it should be in the search results and if so, it adds it back in in the correct sorted order. However, if the course wouldn't be in the search results anyway, then it shouldn't add it back in. This is similar to the behavior that Recal has

frontend/src/Search.js
churichard: It might be cleaner to add a css class or something to a course when it's dragged and dropped into the schedule. Then you can just select for that css class

danielleung27: I'm not sure if it's possible to get the element that had just been dropped using the Dragula library so that's why I have the onclick listener in the update schedule function since that function loops through each course on the schedule. Whenever a course is added it loops through all courses again so I had to do the .off('click') thing so a listener isnt added twice on a course.

churichard: Try the `drop` event: https://github.com/bevacqua/dragula#drakeon-events

tigerpath/views.py
churichard: P.S. I think you can remove this annotation if you add a csrf token to your request. You can try looking here to see how to do it: https://docs.djangoproject.com/en/2.0/ref/csrf/#acquiring-the-token-if-csrf-use-sessions-is-false

frontend/src/Search.js
danielleung27: I'm trying to get the relative ordering back by accessing this.state.data which gives the list of the courses in the search. I think the way React works is that when assigning listeners you have to bind the context of this to the listener so whenever I bind delete onto a button, it only has access to the data at the moment that it was binded. So if I add a cos class, this.state.data will be the list of cos classes and i can only access this list in the listener. If I search ELE and add an ele course, the cos class can't access the current state's data. So I might have to reassign all the listeners whenever a drop event occurs.

frontend/src/Courses.css
```diff
@@ -1,18 +1,95 @@
...
+    width: 120px;
+    height: auto !important;
+    background-color: black;
+    color: #fff;
+    text-align: center;
+    padding: 5px 0;
+    border-radius: 6px;
+    white-space:normal !important;
+ 
+    /* Position the tooltip text - see examples below! */
```
churichard: remove comment

frontend/src/Courses.css
```diff
@@ -1,18 +1,95 @@
...
+  font-size: 13px;
+  padding-left: 2px;
+  width: 100%;
+  height: auto;
+  padding: 5px;
+  background-color: #ffc0cb;
+  list-style-type: none;
+  color: #DB7093;
+/*
+  border-bottom: #ddd;
```
churichard: remove comments

frontend/src/Search.js
```diff
@@ -15,28 +45,44 @@ class Search extends Component {
     };
 
     // select containers to make items draggable
-    let draggableItems = $(".semester").get()
-    draggableItems.push($("#display_courses")[0])
-    let drake = dragula(draggableItems)
+    let draggableItems = $(".semester").get();
+    draggableItems.push($("#display_courses")[0]);
+    let drake = dragula(draggableItems, {
+      copy: function(el, source){return el.parentElement.id == "display_courses";},
```
churichard: Use `===` instead of `==` everywhere

frontend/src/Search.js
```diff
@@ -15,28 +45,44 @@ class Search extends Component {
...
+      accepts: function(el, target){
+        return target.className == "semester";}
+    });
+
+    // check for duplicates to add tooltip, initial num is 3 because 1 from course search, 
+    // 1 from draggable mirror, and 1 in course schedule
+    let addToolTip = function(course){
+      if($('[id=' + course.id +']').length > 3){
```
churichard: `$('[id=' + course.id +']'` can be a variable

frontend/src/Search.js
```diff
@@ -50,29 +96,42 @@ class Search extends Component {
     };
 
     // tells react to post updated course schedule when an item is dropped
-    drake.on('drop', get_courses);
+    drake.on('drop', function(el){
+      // assigns delete listeners to all buttons 
+      // (cant assign to only currently dropped item because copied courses have same ids)
+        $(".delete_course").click(function(){
```
churichard: Is it possible to change the id of a copied course? Might be rather expensive to keep assigning event listeners on every drop event (especially since you're not removing previous ones that have been assigned to the same object)

frontend/src/Search.js
```diff
@@ -50,29 +96,42 @@ class Search extends Component {
...
           if(search_query == this.state.search || search_query == '$')
           {
+            this.setState({data: data});
             ReactDOM.render(
               data.map((course)=> {
-              return <li key={course["id"]} id={course["id"]}>{course["dept"]} {course["number"]}</li>
+              return <li key={course["id"]} id={course["id"]} className='course_display tooltip-custom'>
+              <span className="course_name">{course["listing"]}</span><span className='delete_course'>✖</span><br />
+              <span className='course_title'>{course["title"]}</span>
+              <span className="tooltiptext-custom">Note: class already added</span>
```
churichard: I think this tooltip should only be added to classes that are actually duplicates, not to every class. We may want to add extra tooltips in the future to notify the user of stuff other than duplicate classes

frontend/src/Search.js
```diff
@@ -50,29 +96,42 @@ class Search extends Component {
...
 
   // is called whenever search query is modified
-  update_search(event) {
+  updateSearch(event) {
     ReactDOM.unmountComponentAtNode(document.getElementById('display_courses'));
     this.setState({search: event.target.value});
     let search_query = event.target.value;
     // makes sure that there is always an argument after load_courses, $ is dummy arg
     if(search_query == '') search_query = "$"
+    function removeParent(){this.parentNode.remove();}
```
churichard: What does this do?

frontend/src/Search.js
```diff
@@ -50,29 +96,42 @@ class Search extends Component {
...
         success: function(data) {
-          this.setState({data: data});
           if(search_query == this.state.search || search_query == '$')
           {
+            this.setState({data: data});
             ReactDOM.render(
               data.map((course)=> {
-              return <li key={course["id"]} id={course["id"]}>{course["dept"]} {course["number"]}</li>
+              return <li key={course["id"]} id={course["id"]} className='course_display tooltip-custom'>
+              <span className="course_name">{course["listing"]}</span><span className='delete_course'>✖</span><br />
```
churichard: Minor thing, but I think this delete course button should only appear when a course is added to the schedule

churichard (changes requested):
A few extra things that should be fixed (for reference) (this list supersedes the above list):
- Fix the extra white space that appears after a long course search
- Need to protect the api endpoints somehow (right now anyone can access them, which is a problem)
- Add a django-admin command to combine semester data from past four years into one comprehensive course list (with typical times that the course is offered, typical semester, etc.)

frontend/src/Search.js
danielleung27: It might be a little annoying to do since tooltip should be the last element and I would have to insert the delete element in a specific location rather than using append.

churichard: Better way to do this would be to have some css that positions the delete button instead of having the position be determined by the html. A bit more work but would make styling the courses more flexible

churichard (approved):

PR #18, opened 2018-04-23
Fixed tooltip and whitespace bugs

danielleung27:
tooltip logic makes more sense now ( > 1 item triggers tooltip)


frontend/src/Search.js
```diff
@@ -53,18 +53,25 @@ class Search extends Component {
...
-      let selected_course = $('[id=' + course.id +']');
-      if(selected_course.length > 3){
-        selected_course.each(function(course){
+      let added_courses = $(".semester").find('[id=' + course.id +']');
+      let search_list_course = $("#display_courses").find('[id=' + course.id +']');
+      if(added_courses.length > 1){
+        // add tip to course on search list
+        search_list_course.addClass("showtip");
+        search_list_course.find(".tooltiptext-custom").empty();
+        search_list_course.find(".tooltiptext-custom").append("Note: class already added");
```
churichard: I think you might be able to chain this call after `.empty()`

frontend/src/Search.js
```diff
@@ -53,18 +53,25 @@ class Search extends Component {
...
+        added_courses.each(function(course){
           $(this).addClass("showtip");
+          $(this).find(".tooltiptext-custom").empty();
           $(this).find(".tooltiptext-custom").append("Note: class already added");
```
churichard: I think you might be able to chain this call after `.empty()`

churichard (changes requested):
When you hover over the course in the search list, the tooltip is cut off. I think this happens at certain widths. Can you fix that?
image 2018-04-23 at 11 40 08 am

Might be easier to work with tooltips if you use [Bootstrap's tooltip component](https://getbootstrap.com/docs/4.0/components/tooltips/) instead of making it manually.

danielleung27:
I fixed the tooltips using the old tooltips but ignore it for now, I'm working on the bootstrap tooltip.

frontend/src/Search.js
```diff
@@ -55,26 +55,24 @@ class Search extends Component {
 
     // check for duplicates to add tooltip
     let addToolTip = function(course){
-      let added_courses = $(".semester").find('[id=' + course.id +']');
-      let search_list_course = $("#display_courses").find('[id=' + course.id +']');
+      let added_courses = $('.semester').find('[id=' + course.id +']');
+      let search_list_course = $('#display_courses').find('[id=' + course.id +']');
```
churichard: Forgot to mention this earlier, but css classes should use dashes to separate words (so this should be `display-courses`)

churichard (changes requested):

churichard (approved):

Issue #17, opened 2018-04-23
Course searching issues


churichard:
- Fix the extra white space that appears after a long course search
- Need to protect the api endpoints somehow (right now anyone can access them, which is a problem)
- Add a django-admin command to combine semester data from past four years into one comprehensive course list (with typical times that the course is offered, typical semester, etc.)
- When you add a duplicate course, the tooltip shows "Note: class already added" twice
- Sort by the course listing that is most relevant to what the user typed in (e.g. if the user types in "NEU" it should sort by the NEU course listing from lowest code to highest code - I think right now it just sorts alphabetically)

danielleung27:
Fixed, just made a pull request.

churichard:
#18 fixes some of these issues.

Issues that still need to be addressed:

API:
- Need to protect the api endpoints somehow (right now anyone can access them, which is a problem)

Courses data:
- Scrape courses from past 4 semesters (for now) and store in database
- Add a django-admin command to combine semester data from multiple semesters into one comprehensive course list (with typical times that the course is offered, typical semester, etc.)

Frontend:
- Sort by the course listing that is most relevant to what the user typed in (e.g. if the user types in "NEU" it should sort by the NEU course listing from lowest code to highest code. I think right now it just sorts alphabetically, which means that typing in "NEU" gives you EEB403/NEU403 as the top result. NEU200 as the top result makes more sense)
- Add ability to search by course title

Backend:
- Store user schedule in the database (you can modify the user profile model)

danielleung27:
Update: fixed all issues below in new pull request (except the sorting thing)

For the course searching, I think you can search by course title already. Something might not have shown up because only Spring '18 courses are searchable? For the sorting thing, I think princeton courses and recal have the same schema where the listed courses are sorted alphabetically which is why EEB shows up before NEU. One thing I will fix soon so our search mirrors recal is that the ordering of the cross-listings are off right now. For example, a course is listed "TRA301/COS401/LIN304" on recal but on ours it is listed COS401/LIN304/TRA301. It is actually listed "TRA301/COS401/LIN304" on the course registrar because TRA301 is the primary listing and the rest of the cross-listings are in alphabetical order. I think the django model has a field for primary listing so it should be easy to fix.

[...]

PR #20, opened 2018-04-25
Modified data, schedule save/loading, search all semesters w/ color coding

danielleung27:
Data has dist_area attribute now. Search now displays both semesters with color codes. Api endpoints protected, schedule saves in database and loaded when user logs in.


tigerpath/management/commands/tigerpath_get_courses.py
```diff
@@ -11,12 +11,12 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         scrape_all.get_all_courses()
-        self.stdout.write('course selection: courses scraped successfully')
```
churichard: Might want to keep this line in just so we know the scraper has finished successfully

tigerpath/views.py
```diff
@@ -112,16 +127,23 @@ def filter_courses(search_query):
...
 
+    depts = set(['AAS', 'AFS', 'AMS', 'ANT', 'AOS', 'APC', 'ARA', 'ARC', 'ART', 'ASA', 'AST', 'ATL', 'BCS', 'CBE', 'CEE', 'CGS', 'CHI', 'CHM', 'CHV', 'CLA', 'CLG', 'COM', 'COS', 'CTL', 'CWR', 'CZE', 
+        'DAN', 'EAS', 'ECO', 'ECS', 'EEB', 'EGR', 'ELE', 'ENE', 'ENG', 'ENT', 'ENV', 'EPS', 'FIN', 'FRE', 'FRS', 'GEO', 'GER', 'GHP', 'GLS', 'GSS', 'HEB', 'HIN', 'HIS', 'HLS', 'HOS', 'HPD', 'HUM', 'ISC', 
+        'ITA', 'JDS', 'JPN', 'JRN', 'KOR', 'LAO', 'LAS', 'LAT', 'LCA', 'LIN', 'MAE', 'MAT', 'MED', 'MOD', 'MOG', 'MOL', 'MSE', 'MTD', 'MUS', 'NES', 'NEU', 'ORF', 'PAW', 'PER', 'PHI', 'PHY', 'PLS', 'POL', 
+        'POP', 'POR', 'PSY', 'QCB', 'REL', 'RES', 'RUS', 'SAN', 'SAS', 'SLA', 'SML', 'SOC', 'SPA', 'STC', 'SWA', 'THR', 'TPP', 'TRA', 'TUR', 'TWI', 'URB', 'URD', 'VIS', 'WRI', 'WWS'])
+
     for query in queries:
         if(query == ''):
             continue
         query = query.upper()
```
churichard: You can probably just do `query.upper()` when you're checking the department

tigerpath/views.py
```diff
@@ -112,16 +127,23 @@ def filter_courses(search_query):
         queries = queries + query.split(" ")
 
     # populate with all of semester's courses and convert to course_listings
-    results = models.Semester.objects.get(term_code=max(settings.ACTIVE_TERMS)).course_set.prefetch_related('course_listing_set')
+    # convert semesters to courses
+    results = models.Semester.objects.all()
+    results = models.Course.objects.filter(semester__in=results).prefetch_related('course_listing_set')
     results = models.Course_Listing.objects.filter(course__in=results)
 
+    depts = set(['AAS', 'AFS', 'AMS', 'ANT', 'AOS', 'APC', 'ARA', 'ARC', 'ART', 'ASA', 'AST', 'ATL', 'BCS', 'CBE', 'CEE', 'CGS', 'CHI', 'CHM', 'CHV', 'CLA', 'CLG', 'COM', 'COS', 'CTL', 'CWR', 'CZE',
```
churichard: I actually think Recal's way of doing this (i.e. what you had before) is more maintainable, because you don't have to update the list of depts every time that changes (which, granted, won't be very often, but it still means that someone will have to remember to change this at some point). Like picking your battles, sometimes you have to pick your bugs, and this is probably a bug that no one will really notice (or if they do, it's minor and they won't care very much)

tigerpath/views.py
```diff
@@ -141,12 +163,22 @@ def filter_courses(search_query):
     return models.Course.objects.filter(course_listing_set__in=results)
 
 
-# updates users schedules with added courses
+# updates user's schedules with added courses
```
churichard: If you want to be more precise, `schedules` should probably be `schedule` (or `user's` should be `users'`)

churichard (changes requested):
Make sure to resolve the conflicts with `dev` as well

frontend/src/Search.js
```diff
@@ -157,11 +158,17 @@ class Search extends Component {
...
     // get request, renders list of courses received
-    $.ajax({
+    
+    current_request = $.ajax({
         url: "/api/v1/get_courses/" + search_query,
         datatype: 'json',
         type: 'GET',
         cache: true,
+        beforeSend : function() {           
+          if(current_request != null && current_request.readyState != 4) {
```
churichard: Comment this so readers know 4 means that the request is done

tigerpath/scraper/scrape_all.py
```diff
@@ -7,22 +7,60 @@
...
-            print("----------------------------------")
-            # add dist_area to models
-            for course in scrape_all_courses(term_code):
-                fetch_course = Course.objects.get(registrar_id = str(term_code) + str(course["courseid"]))
-                fetch_course.dist_area = course["area"]
-                fetch_course.save()
-        except Exception as e:
-            raise e
\ No newline at end of file
+     for term_code in term_codes:
```
churichard: this whole for loop should be one space unindented (i.e. what was here before)

tigerpath/scraper/scrape_all.py
```diff
@@ -7,22 +7,60 @@
...
+                 fetch_course.dist_area = course["area"]
+                 fetch_course.save()
+         except Exception as e:
+             raise e
+
+    # update master list
+    all_courses = Course.objects.all()
+    for course in all_courses:
+        # skip if encounter a master course
+        if(course.is_master):
```
churichard: no parentheses in if statements in python

tigerpath/scraper/scrape_all.py
```diff
@@ -7,22 +7,60 @@
...
+            continue
+        fetch_master = all_courses.filter(registrar_id = course.registrar_id[4:])
+        add_semester = ""
+        # update master model if exists
+        if(fetch_master.exists()):
+            # get course from queryset
+            fetch_master = fetch_master[0]
+            # fall semester
+            if(course.registrar_id[3] == '2'):
+                add_semester = "f" + course.registrar_id[1:3]
```
churichard: this is wrong - you should be subtracting the year by 1

tigerpath/scraper/scrape_all.py
```diff
@@ -7,22 +7,60 @@
...
+            # spring semester
+            else:
+                add_semester = "s" + course.registrar_id[1:3]
+            if(add_semester not in fetch_master.semesters):
+                    fetch_master.semesters += " " + add_semester
+            fetch_master.save()
+            # delete duplciate course
```
churichard: fix misspelling

tigerpath/scraper/scrape_all.py
```diff
@@ -7,22 +7,60 @@
...
+            course.delete()
+        else:
+        # convert course to master model
```
churichard: indent

tigerpath/scraper/scrape_all.py
```diff
@@ -7,22 +7,60 @@
...
+            # fall semester
+            if(course.registrar_id[3] == '2'):
```
churichard: this is repeated code - combine it

danielleung27:
The scraper is modified to maintain a master course list and I also modified the course models and scraper to have a field with the string of all of its crosslistings rather than doing database queries to convert from course model to course_listing model which speeds up the searching a lot. Searching for title is not slow anymore (old ajax calls are aborted when new ones are called now) but using abort() throws some errors in the server for some reason. I googled around and it says it might be a server issue so it might not happen on a different server? But searching is still fast but might be faster if this error was fixed.

tigerpath/views.py
```diff
@@ -91,76 +91,61 @@ def update_profile(request, profile_form):
...
+        else:
+            course_info['semester'] = 'spring'
         course_info_list.append(course_info)
-        duplicates[trunc_id] = course.registrar_id
 
+    # sort so courses that match first query show up first (matching num query is not necessary)
+    query = queries[0]
+    # is department
+    if(len(query.upper()) == 3 and query.upper().isalpha()):
+        course_info_list = sorted(course_info_list, key=lambda course: not (course['listing'].startswith(query.upper())))
```
churichard: I think you should also sort by number (100-level classes should appear before 200-level, etc), and it should be the number of the department that the user searched for. If the sorting is stable, you can probably just do it before this line

tigerpath/views.py
```diff
@@ -91,76 +91,61 @@ def update_profile(request, profile_form):
...
+    # check if it matches title
+    else:
+        course_info_list = sorted(course_info_list, key=lambda course: not (course['title'].lower().startswith(query.lower())))
```
churichard: I don't think this is the most natural way to sort - I feel like we shouldn't really care about the placement of the searched word in the title. A better way would be to sort by dept name and code.

tigerpath/views.py
```diff
@@ -91,76 +91,61 @@ def update_profile(request, profile_form):
...
-    results = models.Course.objects.filter(semester__in=results).prefetch_related('course_listing_set')
-    results = models.Course_Listing.objects.filter(course__in=results)
-
-    depts = set(['AAS', 'AFS', 'AMS', 'ANT', 'AOS', 'APC', 'ARA', 'ARC', 'ART', 'ASA', 'AST', 'ATL', 'BCS', 'CBE', 'CEE', 'CGS', 'CHI', 'CHM', 'CHV', 'CLA', 'CLG', 'COM', 'COS', 'CTL', 'CWR', 'CZE', 
-        'DAN', 'EAS', 'ECO', 'ECS', 'EEB', 'EGR', 'ELE', 'ENE', 'ENG', 'ENT', 'ENV', 'EPS', 'FIN', 'FRE', 'FRS', 'GEO', 'GER', 'GHP', 'GLS', 'GSS', 'HEB', 'HIN', 'HIS', 'HLS', 'HOS', 'HPD', 'HUM', 'ISC', 
-        'ITA', 'JDS', 'JPN', 'JRN', 'KOR', 'LAO', 'LAS', 'LAT', 'LCA', 'LIN', 'MAE', 'MAT', 'MED', 'MOD', 'MOG', 'MOL', 'MSE', 'MTD', 'MUS', 'NES', 'NEU', 'ORF', 'PAW', 'PER', 'PHI', 'PHY', 'PLS', 'POL', 
-        'POP', 'POR', 'PSY', 'QCB', 'REL', 'RES', 'RUS', 'SAN', 'SAS', 'SLA', 'SML', 'SOC', 'SPA', 'STC', 'SWA', 'THR', 'TPP', 'TRA', 'TUR', 'TWI', 'URB', 'URD', 'VIS', 'WRI', 'WWS'])
-
+# returns list of courses filtered by query
+def filter_courses(queries):
```
churichard: I think if you take advantage of queryset's lazy evaluation, this function could be even faster: https://docs.djangoproject.com/en/2.0/topics/db/optimization/#understand-querysets. So instead of filtering in Python, you could have the database filter and send back the filtered data

tigerpath/views.py
```diff
@@ -91,76 +91,61 @@ def update_profile(request, profile_form):
...
+    results = models.Course.objects.all()
     for query in queries:
         if(query == ''):
             continue
-        query = query.upper()
 
         # is department
-        if(query in depts):
-            results = list(filter(lambda x: x.dept == query, results))
+        if(len(query.upper()) == 3 and query.upper().isalpha()):
```
churichard: no need to do query.upper() when you're calculating the length and whether or not it's alphabetic 😛

tigerpath/models.py
```diff
@@ -45,6 +45,9 @@ class Course(models.Model):
     description = models.TextField()
     registrar_id = models.CharField(max_length=20)
     dist_area = models.TextField(default="")
+    semesters = models.TextField(default="")
```
churichard: semesters should probably be an [array field](https://docs.djangoproject.com/en/1.11/ref/contrib/postgres/fields/#arrayfield) instead of a string

tigerpath/models.py
```diff
@@ -45,6 +45,9 @@ class Course(models.Model):
...
+    is_master = models.BooleanField(default=False)
+    cross_listings = models.TextField(default="")
```
churichard: cross_listings should probably be an [array field](https://docs.djangoproject.com/en/1.11/ref/contrib/postgres/fields/#arrayfield) instead of a string

tigerpath/views.py
```diff
@@ -91,76 +91,61 @@ def update_profile(request, profile_form):
...
+            results = list(filter(lambda course: query.upper() in course.cross_listings, results))
 
         # is course number
         elif(len(query) <= 3 and query.isdigit() or len(query) == 4 and query[:3].isdigit()):
-            results = list(filter(lambda x: x.number.startswith(query), results))
+            results = list(filter(lambda course: any([course_num.startswith(query) for course_num in re.split('\D+', course.cross_listings)]), results))
```
churichard: This doesn't handle `CEE 102A`

churichard (changes requested):
Another issue: tooltips don't show up when you reload the page.

Also, make sure to resolve conflicts with dev.

tigerpath/views.py
```diff
@@ -106,22 +106,20 @@ def get_courses(request, search_query):
         course_info['id'] = course.registrar_id
         course_info['listing'] = course.cross_listings
         # tag semester
-        if('f' in course.semesters and 's' in course.semesters):
+        if('f' in ''.join(course.all_semesters) and 's' in ''.join(course.all_semesters)):
```
churichard: You should only do the join once. Also, remove parentheses

tigerpath/views.py
```diff
@@ -106,22 +106,20 @@ def get_courses(request, search_query):
...
             course_info['semester'] = 'both'
-        elif('f' in course.semesters):
+        elif('f' in ''.join(course.all_semesters)):
```
churichard: Remove parentheses

tigerpath/views.py
```diff
@@ -106,22 +106,20 @@ def get_courses(request, search_query):
...
-    if(len(query.upper()) == 3 and query.upper().isalpha()):
-        course_info_list = sorted(course_info_list, key=lambda course: not (course['listing'].startswith(query.upper())))
-    # check if it matches title
-    else:
-        course_info_list = sorted(course_info_list, key=lambda course: not (course['title'].lower().startswith(query.lower())))
+    # sort list by dept and code
+    course_info_list = sorted(course_info_list, key=lambda course: course["listing"])
+    # show searched dept first
+    for query in queries:
+        if(len(query) == 3 and query.isalpha()):
```
churichard: Remove parentheses

tigerpath/views.py
```diff
@@ -133,12 +131,12 @@ def filter_courses(queries):
             continue
 
         # is department
-        if(len(query.upper()) == 3 and query.upper().isalpha()):
+        if(len(query) == 3 and query.isalpha()):
```
churichard: Remove parentheses

tigerpath/views.py
```diff
@@ -133,12 +131,12 @@ def filter_courses(queries):
...
             results = list(filter(lambda course: query.upper() in course.cross_listings, results))
 
         # is course number
         elif(len(query) <= 3 and query.isdigit() or len(query) == 4 and query[:3].isdigit()):
```
churichard: Remove parentheses

churichard (changes requested):

churichard (approved):

churichard:
This pull request solves issue #17. Make sure to address the comments above and to resolve the conflicts.

PR #28, opened 2018-05-04
fixed page height

adeniji19:


churichard (approved):

churichard:
Looks good. Should also try working on having separate scroll bars for different semesters

PR #29, opened 2018-05-04
Add course settling

bnehoran:


majors-and-certificates/scripts/verifier.py
```diff
@@ -5,18 +5,15 @@
 import sys
 import collections
 import time
+import copy
+
+from university_info import LANG_DEPTS
 
-schema_location = "schema.json" # path to the requirements JSON schema
 majors_location = "../majors/" # path to folder containing the major requirements JSONs
```
churichard: These should probably be in all caps

majors-and-certificates/scripts/verifier.py
```diff
@@ -36,54 +33,160 @@ def check_major(major_name, courses, year):
...
+    req = _init_req(req)
+    _mark_all_reqs_on_courses(req, courses)
+    _assign_settled_courses_to_reqs(req, courses)
+    _add_course_lists_to_req(req, courses)
+    formatted_courses = _format_courses_output(courses)
+    formatted_req = _format_req_output(req)
+    return formatted_req["satisfied"], formatted_courses, formatted_req
+
+def _init_req(req):
+    req = copy.deepcopy(req)
```
churichard: What's the point of the deepcopy?

majors-and-certificates/scripts/verifier.py
```diff
@@ -36,54 +33,160 @@ def check_major(major_name, courses, year):
...
 
-def _format_major_output(major):
+def _format_req_output(req):
     '''
-    Enforce the type and order of fields in the major output
+    Enforce the type and order of fields in the req output
     '''
     output = collections.OrderedDict()
-    if ("name" not in major) or (major["name"] == '') or (major["name"] == None):
+    if req["name"] == None:
```
churichard: If you wanted to be really pythonic, you can do `if not req["name"]` (this assumes name always exists in req)

majors-and-certificates/scripts/university_info.py
```diff
@@ -0,0 +1,272 @@
...
+    "Technology and Society",
+    "Theater",
+    "Translation and Intercultural Communication",
+    "Urban Studies",
+    "Values and Public Life",
+    "Visual Arts",
+]
+
+# language departments
+LANG_DEPTS = {
```
churichard: You're only using LANG_DEPTS, right? I don't think you should have the other info here (makes it hard to maintain). TBH the LANG_DEPTS shouldn't be here either, it should be in the database.

majors-and-certificates/scripts/verifier.py
```diff
@@ -36,54 +33,160 @@ def check_major(major_name, courses, year):
...
     return output
+    
+def _add_course_lists_to_req(req, courses):
+    include_course_lists = False
+    if "req_list" in req:
+        for subreq in req["req_list"]:
+            if subreq["name"] == None:
+                include_course_lists = True
+            else: # recursively for all named requirements
+                _add_course_lists_to_req(subreq, courses)
```
churichard: The `include_course_lists` variable is a bit janky, can you not just return right after this call?

majors-and-certificates/scripts/tester.py
```diff
@@ -8,7 +8,8 @@
 
 import verifier
 
-tests_location = "verifier_tests"
+schema_location = "schema.json" # path to the requirements JSON schema
```
churichard: These variables should be all caps

churichard (changes requested):
Make sure to also move the `majors-and-certificates` folder into the `tigerpath` folder

majors-and-certificates/scripts/verifier.py
bnehoran: Because it's called as
a = f(b)
so it seems that the intuition would be that the function f doesn't modify b.

majors-and-certificates/scripts/verifier.py
bnehoran: DONE

majors-and-certificates/scripts/verifier.py
bnehoran: Done

majors-and-certificates/scripts/university_info.py
bnehoran: How do I do that without decreasing portability?

majors-and-certificates/scripts/verifier.py
bnehoran: Hmm... on second thought, it makes the code less clear:
Am I checking for a None value? an empty list? an empty string?

majors-and-certificates/scripts/verifier.py
bnehoran: No, because it still has to finish going through the other subrequirements.

majors-and-certificates/scripts/tester.py
bnehoran: DONE

tigerpath/majors-and-certificates/scripts/tester.py
```diff
@@ -8,27 +8,28 @@
 
 import verifier
 
-schema_location = "schema.json" # path to the requirements JSON schema
-tests_location = "verifier_tests" # folder where the test files are stored
+dir_path = os.path.dirname(os.path.realpath(__file__)) # the directory containing this file
```
churichard: This should probably also be in caps unless you're changing it somewhere

tigerpath/majors-and-certificates/scripts/verifier.py
```diff
@@ -53,10 +53,10 @@ def check_degree(degree_name, courses, year):
     :rtype: (bool, dict, dict)
     """
     if degree_name.upper() == "AB":
-        degree_filepath = AB_requirements_location
+        degree_filepath = AB_REQUIREMENTS_LOCATION
     elif degree_name.upper() == "BSE":
-        degree_filepath = BSE_requirements_location
-    return check_requirements(degree_filepath, courses, year)
+        degree_filepath = BSE_REQUIREMENTS_LOCATION
+    return check_requirements(_get_dir_path(),degree_filepath, courses, year)
```
churichard: space after `_get_dir_path(),`

tigerpath/majors-and-certificates/scripts/verifier.py
```diff
@@ -357,18 +361,32 @@ def _mark_settled(path_to, courses):
...
                 num_marked += 1
                 c["reqs_satisfied"].append(path_to)
                 c["settled"].append(path_to)
-                c["used"] = True
     return num_marked
 
+def _check_degree_progress(courses, num_needed, by_semester = None):
+    """
+    Checks whether the correct number of courses have been completed by the 
+    semester number 'by_semester' (1-8)
```
churichard: Just checking, "by semester i" means "by the end of semester i", right? You may want to clarify this in the readme as well.

tigerpath/majors-and-certificates/scripts/verifier.py
```diff
@@ -357,18 +361,32 @@ def _mark_settled(path_to, courses):
...
+    """
+    num_courses = 0
+    for i,sem in enumerate(courses):
```
churichard: Why can't you just do:
```python
for i in range(by_semester):
   num_courses += len(courses[i])
return 1 if num_courses >= num_needed else 0
```
(not tested, adapt as you see fit)

tigerpath/majors-and-certificates/scripts/verifier_tests/1.test.expected
```diff
@@ -165,7 +153,6 @@
     },
     {
       "name": "COS 498",
-      "used": true,
```
churichard: So if a course is not used, then `possible_reqs` is just empty, right?

churichard (changes requested):
Please also document the output of the verifier

majors-and-certificates/scripts/university_info.py
churichard: Difficult to do. Maybe you can keep it as is for now.

tigerpath/majors-and-certificates/scripts/tester.py
bnehoran: DONE

tigerpath/majors-and-certificates/scripts/verifier.py
bnehoran: This was a bug as well. The extra parameter only fits in os.path.join(), not in check_requirements, but I fixed it.

tigerpath/majors-and-certificates/scripts/verifier.py
bnehoran: Done

tigerpath/majors-and-certificates/scripts/verifier.py
bnehoran: Done. Thanks!

tigerpath/majors-and-certificates/scripts/verifier_tests/1.test.expected
bnehoran: It depends on what you mean by "used".

If you want to know whether it was settled and used to actually satisfy a requirement, then look at the length of `reqs_satisfied` (>0 means it was used to satisfy at least one).

On the other hand, if `length(reqs_satisfied)` is 0, the length of `possible_reqs` says how many requirements it *could* be used to satisfy.
If both are 0, then it's really an unneeded course.

That's why I removed that field. It was simultaneously ambiguous and redundant.

churichard (approved):
Remember to also document the output json of the verifier

PR #31, opened 2018-05-05
Display requirements pane

danielleung27:
Added requirement tracker pane


frontend/src/Courses.css
```diff
@@ -43,7 +43,7 @@
 }
 
 li:hover{
-  cursor: pointer;
+  cursor: grab;
```
churichard: I think you might also want to change the cursor so that it is grabbing when the mouse is pressed. So something like this, maybe: 
```css
li:active {
  cursor: grabbing;
}
```

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -7,7 +7,7 @@
 import time
 import copy
 
-from university_info import LANG_DEPTS
+from .university_info import LANG_DEPTS
```
bnehoran: Are you sure about this change?

Adding this period in my code causes the tester code to generate the exception:

```
Traceback (most recent call last):
  File "scripts/tester.py", line 9, in <module>
    import verifier
  File "/home/ubuntu/workspace/tigerpath/majors-and-certificates/scripts/verifier.py", line 10, in <module>
    from .university_info import LANG_DEPTS
SystemError: Parent module '' not loaded, cannot perform relative import
```

bnehoran (dismissed):

tigerpath/static/app/css/app-style.css
```diff
@@ -141,9 +141,10 @@ thead.semester-header {
 #display-courses {
   height: calc(100vh - 95px);
   overflow:auto;
+  background-color: #fff4f4;
```
churichard: I don't think the red tint looks very good, can you remove it?

frontend/src/Requirements.css
```diff
@@ -0,0 +1,68 @@
...
+
+/* count of courses*/
+#requirements .tree-sub-reqs .tree-leaf-req .reqName {
+  font-size:;
+}
+
+#requirements {
+  height: calc(100vh - 60px);
+  overflow: auto;
+  background-color: #fff4f4;
```
churichard: I don't think the red tint looks very good. Maybe try another color. I also don't see what's wrong with white

frontend/src/Search.js
```diff
@@ -46,6 +48,55 @@ class Search extends Component {
...
+            return (<TreeView itemClassName={"tree-leaf-req" + finished}  nodeLabel={reqLabel}>
+              {requirement['settled'].map((course)=>{
+                return(<li className='settled' onClick={(e)=>{toggleSettle(course, requirement['path_to'], false)}}>{course}</li>);
+              })}
+              {requirement['unsettled'].map((course)=>{
+                return(<li className='unsettled text-muted' onClick={(e)=>{toggleSettle(course, requirement['path_to'], true)}}>{course}</li>);
+              })}
+            </TreeView>);
+          }
+        })
```
churichard: This should really be appended to the beginning of the next line, so that it'll look like `}));` in order to align it with the beginning of the block that's being closed

frontend/src/Search.js
```diff
@@ -46,6 +48,55 @@ class Search extends Component {
...
+          if('req_list' in requirement) { 
+            let parentReqLabel = <span>
+                                    <div className='my-arrow'></div>
+                                    {requirement['name']}
+                                 </span>
+            return(<TreeView nodeLabel={parentReqLabel}>{populateReqTree(requirement)}</TreeView>);
+          }
+          else {
+            let finished = '';
+            if(requirement['settled'].length >= requirement['min_needed']) finished=' req_done';
```
churichard: Having the space in front of `req_done` is kind of janky, you may want to put the space after `tree-leaf-req` below instead (I assume that's what the space is for)

frontend/src/Search.js
```diff
@@ -46,6 +48,55 @@ class Search extends Component {
       data: []
     };
 
+    // settles course and runs verifier to update
+    let toggleSettle = function(course, path_to, settle){
+      let courseOnReq = $("#requirements").find("li:contains(\'" + course + "\')");
```
churichard: courseOnReq is never used, do you need this?

frontend/src/Search.js
```diff
@@ -46,6 +48,55 @@ class Search extends Component {
...
+      let courseOnSchedule = $(".semesters").find("p:contains(\'" + course + "\')").parent();
```
churichard: You don't need the backslashes for the single quotes here

frontend/src/Search.js
```diff
@@ -74,6 +126,51 @@ class Search extends Component {
...
+            $.ajax({
+                url: "/api/v1/get_requirements/",
+                datatype: 'json',
+                type: 'GET',
+                cache: true,
+                success: function(data) {
+                  if (data !== null) {
+                    data = data.map((mainReq)=>{
+                      return mainReq[2]
+                    })
```
churichard: You forgot a semicolon

frontend/src/Search.js
```diff
@@ -74,6 +126,51 @@ class Search extends Component {
...
+            // get requirements from existing schedule
+            $.ajax({
+                url: "/api/v1/get_requirements/",
+                datatype: 'json',
+                type: 'GET',
+                cache: true,
+                success: function(data) {
+                  if (data !== null) {
+                    data = data.map((mainReq)=>{
+                      return mainReq[2]
```
churichard: What does `mainReq[2]` mean? Maybe you can add a comment here

tigerpath/majors_and_certificates/scripts/verifier.py
danielleung27: When you run stuff in django, it works out. Individual files in django don't compile/run without errors. Django uses . for relative imports

frontend/src/Search.js
```diff
@@ -134,8 +234,36 @@ class Search extends Component {
       $.ajax({
         url: "/api/v1/update_schedule/",
         type: 'POST',
-        data: courses_taken
+        data: courses_taken,
+        success: function(){
+          // update requirements display
+          // get requirements from existing schedule
+          $.ajax({
```
churichard: This looks like repeated code. Refactor this out into only one place

frontend/src/Search.js
```diff
@@ -134,8 +234,36 @@ class Search extends Component {
...
+                      return <TreeView itemClassName="tree-root" childrenClassName="tree-sub-reqs"nodeLabel={mainReqLabel}>{populateReqTree(mainReq)}</TreeView>
+                    }),
+                    document.getElementById('requirements')
+                  );
+                }
+              }
+          });
+        }
       });
+
```
churichard: Superfluous blank line

frontend/src/Search.js
```diff
@@ -46,6 +48,55 @@ class Search extends Component {
...
+      let courseOnReq = $("#requirements").find("li:contains(\'" + course + "\')");
+      let courseOnSchedule = $(".semesters").find("p:contains(\'" + course + "\')").parent();
+      if(settle){
+        // find course in schedule, attach req path to the course, and update schedule
+        courseOnSchedule.attr('reqs', courseOnSchedule.attr('reqs') + "," + path_to);
+      }
+      else{
+        // find course in schedule, remove req path to the course, and update schedule
+        let pathList = courseOnSchedule.attr('reqs').split(',');
+        let pathListRemoved = courseOnSchedule.attr('reqs').split(',').splice(pathList.indexOf(path_to), 1).join()
```
churichard: This is just
```javascript
pathList.splice(pathList.indexOf(path_to), 1).join();
```

frontend/src/Search.js
```diff
@@ -46,6 +48,55 @@ class Search extends Component {
       data: []
     };
 
+    // settles course and runs verifier to update
+    let toggleSettle = function(course, path_to, settle){
```
churichard: This function (and the general way in which you're handling settling, especially in regards to the `reqs` attribute for the courses) is very confusing to me. Can you explain it?

churichard (changes requested):
A couple of things:

- Move all of the requirements-specific logic outside of Search.js and into a new file
- I think the colors for the requirements pane should be changed - red should mean that a requirement is incomplete, and green should mean that a requirement is complete, and this should be true at all levels of the tree. You can use purple for the courses if you wish
- If you implement the above bullet point, then I think the color for fall semester should be changed as well (and perhaps also the color for fall/spring)
- Requirement headers should also have numbers at all levels (like 1/1)
- Why is degree progress not being checked?
- There's something a little funky with the humanities and social sciences section of the BSE requirements (only 2 is required, and some courses don't immediately settle even though they should). @bnehoran please check this.
- You've mentioned this already, but just to document it here: if a course is settled too quickly, then there are multiple copies of it
- There are some warnings in the console (in Chrome's developer tools) that I think are related to rendering the treeview, try to fix them
- Many of the requirements are 0/0 in Woody Woo. If min_needed is null maybe you should hide this number instead of displaying /0. This also underscores why we need to have numbers at all levels of the requirement headers

frontend/src/Courses.css
danielleung27: This makes the grabbing cursor only if you are move within the list item and turns into other cursors if you move outside. Leaving this here to fix later.

frontend/src/Search.js
danielleung27: screenshot_1
Does this look better?

frontend/src/Search.js
danielleung27: Whenever a course is rendered on the requirements pane, it has a class 'settled' or 'unsettled text-muted' for styling settled and unsettled classes. A toggle settle function is also attached to these courses.

The courses that are displayed on the semesters is what is sent back through to the verifier so the toggleSettle function attaches more information to these elements so they are sent back as well so the backend (verifier) has access to it. The verifier takes in a list of paths (each requirement has a path for example: 'Computer Science - BSE // Core Courses // Theory') so it knows which requirement the course is supposed to satisfy. I attach these paths to the 'req' attribute of the courses on the schedule. The verifier reads this list and adds this course to each path requirement's 'settled' list. toggleSettle will remove the path and notify the verifier to update if the requirement being clicked on is already settled (the if statement handles this).

frontend/src/Search.js
churichard: I wasn't pointing to that part, I was looking below that. Line 95 closes line 70 from what I can tell, so it should be aligned with it. Something like this:
fbf3faa9c5c31a0c27a03b41ed0b271e _image 202018-05-06 20at 2012 21 33 20am
In JavaScript it's just generally very easy to misalign curly braces and parentheses, which makes the code harder to read. It's important to make sure that things are aligned properly

danielleung27:
I think I fixed everything except the bug where you settle a course multiple times and also for some reason the grabbing cursor stops grabbing when you move outside of the list item in the search.

Actually I may make another push, I might be able to implement the degree progress tracker.
update: done

tigerpath/majors_and_certificates/scripts/verifier.py
bnehoran: Well, if you're right, then isn't that a problem if the testing script cannot run?

churichard: @danielleung27 @bnehoran  Try `scripts.university_info` instead?

bnehoran: Okay @danielleung27 , you are right.

`from .university_info import LANG_DEPTS`
is the right way to do it since it's treating the scripts folder as a package.

The fix was just to move the tester.py file out of the package so that it can reference it, and do 
`from scripts import verifier`

I'll create a different pull request for that. No need for you to worry about it.

I approve then. Sorry about that.

churichard:
Before I look at this again, please update this branch with the latest changes in dev and resolve the conflicts

frontend/src/Requirements.js
```diff
@@ -0,0 +1,96 @@
...
+        let parentReqLabel = <span>
+                                  <div className='my-arrow'></div>
+                                  <span className='reqName'>{requirement['name']}</span>
+                                  <span className='reqCount'>{tag}</span>
+                               </span>
+        return(<TreeView key={treeHash} nodeLabel={parentReqLabel} itemClassName={finished}>{populateReqTree(requirement)}</TreeView>);
+      }
+      else {
+        // treeview key is not needed but assigning here to prevent error in console, this function creates a hash from the name
+        let treeHash = requirement['name'].split("").reduce(function(a,b){a=((a<<5)-a)+b.charCodeAt(0);return a&a},0);
```
churichard: Please refactor all of this duplicate code

frontend/src/Requirements.js
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// traverses req tree to display when updating reqlist
+export function populateReqTree(reqTree){
+  return(reqTree['req_list'].map((requirement)=>{
+      if('req_list' in requirement) { 
+        // treeview key is not needed but assigning here to prevent error in console, this function creates a hash from the name
+        let treeHash = requirement['name'].split("").reduce(function(a,b){a=((a<<5)-a)+b.charCodeAt(0);return a&a},0);
+        let finished = '';
+        if((requirement['min_needed'] == 0 && requirement['count'] >= requirement['max_counted']) ||
```
churichard: Please always use === instead of ==

frontend/src/Search.js
```diff
@@ -38,6 +44,99 @@ $.ajaxSetup({
...
+    success: function(){
+      // update requirements display
+      // get requirements from existing schedule
+      $.ajax({
+          url: "/api/v1/get_requirements/",
+          datatype: 'json',
+          type: 'GET',
+          cache: true,
+          success: function(data) {
+            console.info(data)
```
churichard: Useful for debugging but shouldn't be in production

churichard (changes requested):
- Should show the requirements even if the user_schedule is null (which means that the user is a new user with a blank schedule)
- If you take the same course twice, it should show up as unsettled twice in the requirements. Right now it shows up as unsettled once, and then when you settle it, it shows up as settled twice.

bnehoran:
@churichard the "only appearing as settled once" is actually my fault from the verifier. I had the verifier remove duplicates from the unsettled list, but I could fix that.

On the other hand, the problem with being settled twice is probably because the front end "settles" all the courses of the same kind (I don't know if determined by name, or ID, or what not) all at once.
So Daniel, see if you can fix that.

frontend/src/Requirements.js
```diff
@@ -9,32 +9,44 @@ import {updateSchedule} from './Search';
 
 // settles course and runs verifier to update
 export function toggleSettle(course, path_to, settle){
-  let courseOnSchedule = $(".semesters").find("p:contains('" + course + "')").parent();
+  let allCoursesOnSchedule = $(".semesters").find("p:contains('" + course + "')").parent();
+  let courseOnScheduleAssigned = '';
+  let courseOnScheduleNotAssigned = '';
+  // checks for duplicates on course schedule and grabs courses that do not have the path and courses that do have the path
+  allCoursesOnSchedule.each(function(){
+    if($(this).attr('reqs').indexOf(path_to) === -1) courseOnScheduleNotAssigned = $(this);
```
churichard, line 17: Does this work if there are 3+ duplicate courses? Also, I get strange behavior where once I settle a course, another duplicate course shows up as unsettled right below it. I'm gonna say this is okay for now if you can't fix it, but it's something that needs to be fixed as part of issue #41

frontend/src/Requirements.js
```diff
@@ -9,32 +9,44 @@ import {updateSchedule} from './Search';
...
+  return stringName.split("").reduce(function(a,b){a=((a<<5)-a)+b.charCodeAt(0);return a&a},0);
+}
+
 // traverses req tree to display when updating reqlist
 export function populateReqTree(reqTree){
   return(reqTree['req_list'].map((requirement)=>{
       if('req_list' in requirement) { 
         // treeview key is not needed but assigning here to prevent error in console, this function creates a hash from the name
-        let treeHash = requirement['name'].split("").reduce(function(a,b){a=((a<<5)-a)+b.charCodeAt(0);return a&a},0);
+        let treeHash = getHash(requirement['name'])
```
churichard: It's not just this line but also the next 13 or so lines that are repeated between this if block and the next else block. Please refactor that out

frontend/src/Search.js
```diff
@@ -38,6 +44,99 @@ $.ajaxSetup({
...
+  });
+  courses_taken = JSON.stringify(courses_taken);
+  $.ajax({
+    url: "/api/v1/update_schedule/",
+    type: 'POST',
+    data: courses_taken,
+    success: function(){
+      // update requirements display
+      // get requirements from existing schedule
+      $.ajax({
```
churichard: Factor out this ajax call into another function, maybe `renderRequirements`

frontend/src/Search.js
```diff
@@ -75,6 +177,8 @@ class Search extends Component {
               updateSchedule();
             });
           }
+          // get requirements from existing schedule
+          updateSchedule();
```
churichard: Should call `renderRequirements` instead of `updateSchedule` after you make the change above

churichard (changes requested):

frontend/src/Requirements.js
bnehoran: Does this happen regularly? Or only when you have duplicate courses?

churichard: Just for duplicate courses I believe

churichard (approved):

PR #32, opened 2018-05-05
Add ELE major

bnehoran:


danielleung27:
For bse major, the humanities counter doesn't work perfectly. As mentioned before min_needed is only 2. Also I noticed that it only counts the first 4 courses that satisfy the 4 areas in bse (EC, LA, etc). So if you have 2 LAs, only 1 gets counted. Also the courses in the humanities don't auto-settle.

churichard:
I think this pull request got polluted - you may want to update it with the latest changes from dev

PR #35, opened 2018-05-07
Add support for double-counting courses

bnehoran:


tigerpath/majors-and-certificates/scripts/verifier.py
```diff
@@ -167,7 +168,14 @@ def _add_course_lists_to_req(req, courses):
...
         for sem in courses:
             for course in sem:
-                if len(course["settled"]) > 0:
+                if req["double_counting_allowed"]: 
+                    if len(course["reqs_double_counted"]) > 0:
+                        for path in course["reqs_double_counted"]:
+                            if req["path_to"] in path:
+                                req["settled"].append(course["name"])
+                                course["reqs_satisfied"].append(req["path_to"]) ## add to these lists because couldn't
+                                course["settled"].append(req["path_to"]) ## be added in _assign_settled_courses_to_reqs()
```
churichard: You're adding the path to `settled` as well?

tigerpath/majors-and-certificates/scripts/verifier.py
```diff
@@ -167,7 +168,14 @@ def _add_course_lists_to_req(req, courses):
         req["unsettled"] = []
         for sem in courses:
             for course in sem:
-                if len(course["settled"]) > 0:
+                if req["double_counting_allowed"]: 
+                    if len(course["reqs_double_counted"]) > 0:
+                        for path in course["reqs_double_counted"]:
+                            if req["path_to"] in path:
```
churichard, line 174: Why `in` and not `==`?

churichard (changes requested):

tigerpath/majors-and-certificates/scripts/verifier.py
bnehoran: Should I not?

tigerpath/majors-and-certificates/scripts/verifier.py
bnehoran: Because this only runs through named requirements, it might only arrive at a parent of the requirement actually satisfied.

For example, if a course satisfied the requirement
`"Electrical Engineering//Senior Thesis//001"`,
then the code will only ever get to 
`"Electrical Engineering//Senior Thesis"`,
which is where it should then place the class.

So it just checks for prefix match.
Technically, `in` checks for substring match, but unless one of the subrequirements is named exactly like the major, it's essentially the same.

bnehoran: Should I switch to a more explicit prefix match function instead?

tigerpath/majors-and-certificates/scripts/verifier.py
```diff
@@ -371,6 +393,19 @@ def _mark_settled(path_to, courses):
                 c["settled"].append(path_to)
```
bnehoran: Need to correct 
`elif len(c["possible_reqs"]) == 1`
to check if course can only satisfy one no-double-counting requirement.

bnehoran: Done!

tigerpath/majors-and-certificates/scripts/verifier.py
bnehoran: Removed

tigerpath/majors-and-certificates/scripts/verifier.py
churichard: Nah, this is good

tigerpath/majors-and-certificates/scripts/verifier.py
```diff
@@ -191,8 +191,9 @@ def _init_courses(courses, req_name = None):
             course["name"] = course["name"].split(':')[0]
             course["possible_reqs"] = []
             course["reqs_satisfied"] = []
-            course["reqs_double_counted"] = []
+            course["reqs_double_counted"] = [] # reqs satisfied for which double counting allowed
             course["semester_number"] = sem_num
+            course["num_settlable"] = 0 # number of reqs to which can be settled. autosettled if 1
```
churichard: should be `num_settleable`

churichard (changes requested):

churichard (approved):

churichard:
Make the change above before you merge

PR #36, opened 2018-05-07
Moved the tester code

bnehoran:


tigerpath/majors-and-certificates/tester.py
```diff
@@ -26,7 +26,7 @@ def main():
                 year = int(f.readline())
                 courses = json.loads(f.read())
             major_filename = major_name + "_" + str(year)  + ".json"
-            major_filepath = os.path.join(DIR_PATH, verifier.MAJORS_LOCATION, major_filename)
+            major_filepath = os.path.join(DIR_PATH, "scripts", verifier.MAJORS_LOCATION, major_filename)
```
churichard, line 29: This is a little strange to me (you're going into the scripts folder and then coming out of the scripts folder (since MAJORS_LOCATION is "../majors/"). I think what might be better is if you move the verifier.py and university_info.py out of the scripts folder and just into the root level of majors-and-certificates - then you won't need the ".." to jump up a directory

churichard (changes requested):

tigerpath/majors-and-certificates/tester.py
bnehoran: The problem is just that the testing code apparently has to always be a level higher than the code it imports so that python will treat the import as a package (and then you can use relative imports like `from .university_info import LANG_DEPTS`). If the testing code is *within* the package, it seems that it cannot be run directly:
https://stackoverflow.com/questions/14132789/relative-imports-for-the-billionth-time

If I move the verifier code to the `majors_and_certificates` folder, would I then move the tester to the `tigerpath` folder?

churichard: What you could try doing is renaming the `scripts` folder to `verifier` and then moving the `majors`, `degrees`, and `certificates` folders inside `verifier`

churichard: Or what might be even better is if the verifier doesn't hardcode the paths to the majors, degrees, and certificates at all - it's just passed in as part of the file_name argument to the verifier

I think this is better because the verifier should be abstracted away from the location of the majors, degrees, and certificates. It just handles the verifying, it shouldn't care about where the files are located - that's the job of whoever is calling the verifier.

But I guess you could also argue that these files are part of the "package" of the verifier, in which case I think you should do what I said above.

PR #43, opened 2018-05-08
Added grip dots to courses

danielleung27:


danielleung27:
screenshot_1
screenshot_2

Which one looks better?

churichard:
Try different dots (maybe from the FontAwesome library).

Also, I think that the dots should be closer to the left margin and farther away from the text.

danielleung27:
screenshot_4
screenshot_3

churichard:
I like the first image better, where the dots run down the width of the card. Maybe you can try making the dots a little bit smaller

danielleung27:
screenshot_5
I also reduced the opacity

churichard:
Grip dots have been superseded by #96

Issue #44, opened 2018-05-08
Bug: Courses disappearing from requirements tree


bnehoran:
Courses disappear whenever they are settled in requirements whose names contain commas.

One such example is
`"Woodrow Wilson School of Public and International Affairs//Prerequisites//Politics, Sociology or Psychology Prerequisite"`
Try settling `SOC222` in it, for example.

Or try settling `AST309 / ENE309 / MAE309 / PHY309` in 
`"Woodrow Wilson School of Public and International Affairs//Core//Elective Courses//Security and Sustainability//Environment, Climate, and Energy"`

I believe this is because the settled list is stored as a comma separated string.
Either find a way to store an actual list, or escape the identifiers of requirements before adding them (and unescape when reading).

danielleung27:
Bug fixed

PR #47, opened 2018-05-08
Fix bug when path has commas

danielleung27:
#44 Changed delimiter from ',' to '$'

### DIFF
--- a/frontend/src/Search.js
+++ b/frontend/src/Search.js
@@ -180,7 +180,7 @@ class Search extends Component {
             });
           }
           // get requirements from existing schedule
-          renderRequirements();
+          updateSchedule();
         }
     });
 
--- a/frontend/src/Search.js
+++ b/frontend/src/Search.js
@@ -110,9 +110,9 @@ export function updateSchedule(){
   let added_courses = document.querySelectorAll(".semester");
   let courses_taken = [];
   let i = 0;
-  added_courses.forEach(function (semester){
+  added_courses.forEach(function (semester, sem_num){
     courses_taken.push([]);
-    semester.childNodes.forEach(function(course){
+    semester.childNodes.forEach(function(course, course_index){
       if(typeof course.innerHTML !== 'undefined'){
         let course_entry = {}
         course_entry["name"] = course.getElementsByClassName("course-name")[0].innerHTML;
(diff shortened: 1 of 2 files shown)


churichard (changes requested):
This is janky, look at https://stackoverflow.com/a/432353/1744102

frontend/src/Search.js
```diff
@@ -121,7 +121,7 @@ export function updateSchedule(){
         course_entry["semester"] = course.className.split(" ")[1];
         course_entry["dist_area"] = course.getAttribute('dist_area');
         // slice the last element out because it's an empty string: format is "x,y,z,"
```
churichard: Change comment - it's not relevant anymore, right?

frontend/src/Search.js
```diff
@@ -121,7 +121,7 @@ export function updateSchedule(){
...
-        course_entry["settled"] = course.getAttribute('reqs').split(',').slice(0, -1);
+        course_entry["settled"] = $('.semester').slice(sem_num, sem_num + 1).find('li').slice(course_index, course_index + 1).data('reqs');
```
churichard: Why are you using slice - can't you just index into the array?

frontend/src/Search.js
```diff
@@ -158,16 +158,15 @@ class Search extends Component {
...
-                if(course['settled'].length > 0) courseReqs = course['settled'].join(',') + ',';
-                return <li key={course["id"]} id={course["id"]} className={"course-display " + course["semester"]} dist_area={course["dist_area"]} reqs={courseReqs}>
+                return <li key={course["id"]} id={course["id"]} className={"course-display " + course["semester"]} dist_area={course["dist_area"]}>
                 <p className="course-name">{course["name"]}</p><i className="fas fa-times-circle delete-course"></i>
                 <p className="course-title">{course["title"]}</p>
                 </li>
               }), document.getElementById('sem' + index))
-              semester.map((course) => {
+              semester.map((course, course_index) => {
+                $('.semester').slice(sem_num, sem_num + 1).find('li').slice(course_index, course_index + 1).data('reqs', course['settled'])
```
churichard: Why are you using slice - can't you just index into the array?

churichard (changes requested):

frontend/src/Search.js
danielleung27: Indexing into the array was giving me the html element rather than a Jquery selection so I couldn't use .find on it, slicing returns a jquery selection so .find would work. .get() was also giving me an html element. I couldn't find a function to get from index and return a Jquery selection.

churichard: Try `.eq`: https://api.jquery.com/eq/

frontend/src/Search.js
```diff
@@ -181,7 +179,7 @@ class Search extends Component {
             });
           }
           // get requirements from existing schedule
-          renderRequirements();
+          updateSchedule();
```
churichard, line 182: Why did you change this?

churichard (changes requested):

frontend/src/Search.js
danielleung27: This fixes the null schedule problem. On startup it should run updateSchedule function so that even if the schedule is empty, it sends back an empty list and sets the user_schedule to equal that empty list. Then the requirements can display. If the schedule is null, the requirements function can't run properly.

churichard (approved):

PR #69, opened 2018-05-11
Change user profile major to foreign key

churichard:
This is so that the production website's database will migrate properly

Also fixed a minor annoyance where the text in a course card sometimes moved to the next line on hover (because of the appearance of course info/eval buttons on the right)


tigerpath/models.py
```diff
@@ -3,9 +3,9 @@
 from django.dispatch import receiver
 from django.conf import settings
 from django.contrib.auth.models import User
-from django.contrib.postgres.fields import JSONField
+from django.contrib.postgres.fields import JSONField, ArrayField
+
```
bnehoran, line 7: Why the extra line?

tigerpath/models.py
```diff
@@ -155,7 +155,7 @@ class UserProfile(models.Model):
                                 on_delete=models.CASCADE,
                                 related_name='profile')
     nickname = models.CharField(max_length=50, null=True, blank=True)
-    major = models.CharField(max_length=7, null=True)
+    major = models.ForeignKey(Major, related_name="+", on_delete=models.CASCADE, null=True)
```
danielleung27, line 158: Why is related_name="+"

danielleung27 (approved):

tigerpath/models.py
churichard: This means that majors don't link back to user profiles

tigerpath/models.py
churichard: It's convention to put an extra line between "from" imports and standard imports

PR #71, opened 2018-05-11
Add collapsed course lists and additional info to verifier output

bnehoran:


tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -465,6 +485,47 @@ def _course_match(course_name, pattern):
                 return True
     return False
 
+def _get_req_by_path(req, path_to):
```
churichard, line 488: Where is this supposed to be used? It's not a public function and isn't used anywhere in your code.

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -465,6 +485,47 @@ def _course_match(course_name, pattern):
...
+    Returns sets of all courses and of all distribution requirements
+    in req's subtree.
+    Note: Sets may contain duplicate courses if a course is listed in multiple
+    different ways
+    '''
+    if "course_list" in req:
+        course_set = set()
+        for course in req["course_list"]:
+            course = course.split(':')[0] # strip course name
+            # course = "".join(course.split()).upper() # remove spaces
```
churichard: Remove this comment

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -465,6 +485,47 @@ def _course_match(course_name, pattern):
...
+        return req
+    if "req_list" in req:
+        for subreq in req["req_list"]:
+            if _get_req_by_path(subreq, path_to):
+                return subreq
+    return None
+
+def _get_collapsed_course_and_dist_req_sets(req):
+    '''
+    Returns sets of all courses and of all distribution requirements
```
churichard: Say it's a tuple of (the set of all courses, the set of all distribution requirements)

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -465,6 +485,47 @@ def _course_match(course_name, pattern):
...
+        return (course_set, set())
+    if "dist_req" in req:
+        return (set(), set([req["dist_req"]]))
+    total_course_set = set()
+    total_dist_req_set = set()
+    if "req_list" in req:
+        for subreq in req["req_list"]:
+            course_set, dist_req_set = _get_collapsed_course_and_dist_req_sets(subreq)
+            if course_set:
+                total_course_set |= course_set
```
churichard: Might want to make a comment saying this is taking the union of both sets

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -465,6 +485,47 @@ def _course_match(course_name, pattern):
...
+        _init_path_to(req)
+    if req["path_to"] == path_to:
+        return req
+    if "req_list" in req:
+        for subreq in req["req_list"]:
+            if _get_req_by_path(subreq, path_to):
+                return subreq
+    return None
+
+def _get_collapsed_course_and_dist_req_sets(req):
```
churichard, line 502: Do you think this might be better as a public function instead of integrated into check_requirements? This information is only relevant when the user wants to know what courses satisfy a requirement. It makes more sense to only return that information when necessary, instead of every time when the user settles a course.

churichard (changes requested):

tigerpath/majors_and_certificates/scripts/verifier.py
bnehoran: Done

tigerpath/majors_and_certificates/scripts/verifier.py
bnehoran: Done

tigerpath/majors_and_certificates/scripts/verifier.py
bnehoran: I thought this should be clear / self evident.

It would be almost like writing
`i +=1 # increment the variable i`

bnehoran: But okay, I added the comment

churichard (approved):

PR #77, opened 2018-05-13
Update docker to use pipfile instead of requirements

churichard:
This is so our python dependencies aren't being maintained across two different files (pipfile and requirements.txt)


churichard:
@adeniji19 can you test this and make sure your docker still works

PR #80, opened 2018-05-13
Added loading spinner for search

danielleung27:


churichard (changes requested):
Already mentioned these things, but putting them here for reference:
- Try a fontawesome icon instead of the gif (I recommend fa-spinner): https://www.w3schools.com/icons/fontawesome_icons_spinner.asp
- Try using the old search results text
- Center the spinner with the results text

churichard:
Also, searching for a "#" makes the loading spinner appear forever.

churichard (approved):

PR #82, opened 2018-05-13
Master course updates info with latest course info

danielleung27:


tigerpath/scraper/scrape_all.py
```diff
@@ -61,12 +61,18 @@ def get_all_courses():
         if fetch_master.exists():
             # get course from queryset
             fetch_master = fetch_master[0]
-            
             if add_semester not in fetch_master.all_semesters:
                 fetch_master.all_semesters.append(add_semester)
+            # update all information of master with new course info
+            fetch_master.title = course.title
+            fetch_master.rating = course.rating
```
churichard: rating is not used, no need to update

tigerpath/scraper/scrape_all.py
```diff
@@ -61,12 +61,18 @@ def get_all_courses():
...
+            fetch_master.description = course.description
+            fetch_master.dist_area = course.dist_area
```
churichard: I'm pretty sure dist_area shouldn't change. It's kind of problematic to update it like this because then one course might fulfill different dist_areas depending on what semester it was taken in. Let's just assume this doesn't happen

churichard (changes requested):
So when the course number changes but the registrar id is the same, this doesn't do anything with the course number, right? I think it's fine for now, but we should think about what the best way to handle updated courses is

tigerpath/management/commands/tigerpath_get_courses.py
```diff
@@ -11,12 +10,4 @@ class Command(BaseCommand):
...
-        self.stdout.write('course selection: courses scraped successfully')
-#        for term_code in settings.ACTIVE_TERMS:
-#            results = get_courses_by_term_code(term_code)
-#            data = json.dumps(results)
-#            caches['courses'].set(term_code, data)
-#            self.stdout.write(
-#                'course selection: cache regenerated for term ' + str(term_code))
-#        caches['courseapi'].clear()
-#        self.stdout.write('course selection: courseapi cache cleared.')
+        self.stdout.write('course selection: courses scraped successfully')
```
churichard: Files should have newlines at the end

tigerpath/views.py
```diff
@@ -262,69 +262,4 @@ def get_requirements(request):
...
-        'title': course.title,
-        'sections': sections,
-        'semester': hydrate_semester(course.semester),
-    }
-
-
-def get_courses_by_term_code(term_code):
-    filtered = Course.objects.filter(Q(semester__term_code=term_code))
-    return [hydrate_course_dict(c) for c in filtered]
+    return HttpResponse(ujson.dumps(requirements, ensure_ascii=False), content_type='application/json')
```
churichard: Files should have newlines at the end

tigerpath/scraper/scrape_all.py
```diff
@@ -62,16 +63,13 @@ def get_all_courses():
             # get course from queryset
             fetch_master = fetch_master[0]
             if add_semester not in fetch_master.all_semesters:
```
churichard: Just wondering, are there cases where add_semester would actually already be in fetch_master.all_semesters?

tigerpath/scraper/scrape_all.py
```diff
@@ -62,16 +63,13 @@ def get_all_courses():
...
-                fetch_master.all_semesters.append(add_semester)
-            # update all information of master with new course info
-            fetch_master.title = course.title
-            fetch_master.rating = course.rating
-            fetch_master.description = course.description
-            fetch_master.dist_area = course.dist_area
-            fetch_master.cross_listings = course.cross_listings
-            fetch_master.save()
+                course.all_semesters = deepcopy(fetch_master.all_semesters)
+                if add_semester not in course.all_semesters:
```
churichard: This is always true

tigerpath/scraper/scrape_all.py
```diff
@@ -62,16 +63,13 @@ def get_all_courses():
...
+                    course.all_semesters.append(add_semester)
+            course.is_master = True
```
churichard: You forget to strip off the first 4 characters of the registrar_id. Might be better to move 
```python
# strip semester off id
course.registrar_id = course.registrar_id[4:]
course.is_master = True
course.save()
```
after this if block and the next else block, since it's in both.

churichard (changes requested):

churichard:
Also, in this latest code change, when the course number changes, doesn't the scraper just nuke the previous course number? Do you think it might be better to copy the course?

Actually, maybe that would be too complicated, because then we would have two courses with the same registrar id. Maybe we should have different cross listings based on the semester the course was offered in.

Or we can just assume that course number changes don't happen very often, and when it does, a student would know about the new course number. This is probably the simplest way to handle things, because it would require no change on your end.

tigerpath/scraper/scrape_all.py
danielleung27: Just in case we rescrape a semester it wont add a semester twice.

danielleung27:
I think the 3rd option would be the best.

churichard (approved):

PR #84, opened 2018-05-13
Search fix special characters (loading spinner shows forever)

danielleung27:
Haven't fixed the / yet but search query is encoded now so should support most special characters. Slashes seem to be a special case
https://code.djangoproject.com/ticket/15718


churichard:
Can you merge with dev? I think there's extra code here

churichard (changes requested):
Merge with dev

churichard (commented):
Update this to work with slashes

churichard:
This is a pretty old PR. I'm gonna close it because I think eventually we're going to want to move to using a query param for the search query, which will fix these issues.

PR #85, opened 2018-05-13
Added Princeton Shield Favicon

adeniji19:


churichard:
I think we should have our own logo.

PR #91, opened 2018-05-14
Instruct the user to search during tutorial step so they can see the search results in the next step.

bnehoran:
A simple way to make the tutorial more effective by allowing users to actually see the search results while the search results are being explained, since showing is better than telling.

Try it out!


adeniji19 (approved):
This is a great idea

churichard:
Should say "press enter, then click next" to indicate to the user that they should click next to advance the tutorial

churichard (approved):

PR #96, opened 2018-05-15
Add course drop shadow, add padding to search results

churichard:
The drop shadow signifies that the course cards are draggable. The padding in the search results is so that the drop shadow looks better and so that the UI is more uniform (now everything looks like cards)


churichard:
Here's what it looks like (this is when the cursor is clicking on COS 126)
screen shot 2018-05-15 at 1 08 02 pm

adeniji19 (approved):

PR #97, opened 2018-05-15
Add rounded corners to course cards

churichard:
Very minor UI change. Made for consistency between the courses in the search results, schedule, and requirements pane

### DIFF
--- a/tigerpath/templates/tigerpath/index.html
+++ b/tigerpath/templates/tigerpath/index.html
@@ -83,8 +83,6 @@
         </div>
         <div id="display-courses">
         </div>
-
-        <!-- Requirements -->
       </div>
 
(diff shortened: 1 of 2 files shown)


tigerpath/templates/tigerpath/index.html
```diff
@@ -83,8 +83,6 @@
         </div>
         <div id="display-courses">
         </div>
-
-        <!-- Requirements -->
```
bnehoran, line 87: Is this a remnant from something that was removed earlier?

bnehoran (approved):

tigerpath/templates/tigerpath/index.html
churichard: I think the requirements were probably there originally but got moved elsewhere

PR #106, opened 2018-05-17
Add mobile banner

### LINKED ISSUE #95: Show banner on mobile
When a user uses TigerPath on mobile, it should show a banner saying that the experience is optimized for desktops/laptops

churichard:
Shows up when the screen size is small (fixes #95)
image 2018-05-16 at 10 20 16 pm

### DIFF
--- a/tigerpath/static/app/js/mobile_banner.js
+++ b/tigerpath/static/app/js/mobile_banner.js
@@ -1,7 +1,0 @@
-window.addEventListener("resize", function() {
-  if (window.innerWidth <= 991) {
-    $('#banner').show();
-  } else {
-    $('#banner').hide();
-  }
-});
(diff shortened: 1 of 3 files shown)


tigerpath/static/app/js/mobile_banner.js
```diff
@@ -0,0 +1,7 @@
+window.addEventListener("resize", function() {
+  if (window.innerWidth <= 991) {
```
bnehoran: Should `991` really be hardcoded or is there some existing parameter that is used by Bootstrap that can be read?
Either way, it should probably be factored out.

bnehoran (changes requested):

tigerpath/static/app/js/mobile_banner.js
churichard: You're right, the JavaScript code is superfluous. Everything is now done in css

churichard: In the CSS, I'm comparing using a fractional width, which is standard for Bootstrap media queries: https://getbootstrap.com/docs/4.1/layout/overview/

Unfortunately CSS doesn't have variables, so one thing we can look into adopting in the future is Sass

bnehoran (approved):

PR #116, opened 2018-05-27
Search by req

danielleung27:
Added search by req button to requirement nodes (not fully styled, will attach onto a popover in a new pr)


frontend/src/Search.js
```diff
@@ -242,65 +296,13 @@ class Search extends Component {
...
-            )
-            ReactDOM.render(<span id='search-count-num'>{data.length}</span>, document.getElementById('search-count'))
+            ReactDOM.render(returnSearchList(data), document.getElementById('display-courses'));
+            ReactDOM.render(<span id='search-count-num'>{data.length}</span>, document.getElementById('search-count'));
             $('#spinner').css('display', 'none');
           }
         }.bind(this)
     });
 
-    /* Converts semester to term code */
```
danielleung27, line 265: Moved functions outside so returnSearchList will have access to it. I also moved returnSearchList outside because requirements.js needed access to that function.

frontend/src/Requirements.js
```diff
@@ -15,8 +16,8 @@ export function toggleSettle(course, path_to, settle){
   // checks for duplicates on course schedule and grabs courses that do not have the path and courses that do have the path
   allCoursesOnSchedule.each(function(){
     if($(this).data('reqs').map((path)=>{
-      return path.split('//')[0]
-    }).indexOf(path_to.split('//')[0]) === -1) courseOnScheduleNotAssigned = $(this);
+      return path.split('//')[2]
```
bnehoran: The path identifier format now starts with the path prefix `type//year//code_or_name`, and so it should always compare the full prefix consisting of all three elements, not just the third.

frontend/src/Requirements.js
```diff
@@ -15,8 +16,8 @@ export function toggleSettle(course, path_to, settle){
...
+    }).indexOf(path_to.split('//')[2]) === -1) courseOnScheduleNotAssigned = $(this);
```
bnehoran: Same as above

frontend/src/Requirements.js
```diff
@@ -33,6 +34,21 @@ function getHash(stringName) {
...
+  $.ajax({
+      // the slashes messes up the url 
+      url: "/api/v1/get_req_courses/" + req_path.replace(/\/\//g, '$'),
+      datatype: 'json',
+      type: 'GET',
+      cache: true,
+      success: function(data) {
+        ReactDOM.render(returnSearchList(data), document.getElementById('display-courses'));
+        ReactDOM.render(<span id='search-count-num'>{data.length}</span>, document.getElementById('search-count'));
+        $('#spinner').css('display', 'none');
```
bnehoran, line 49: Why doesn't the spinner show while loading the courses for a requirement?

frontend/src/Requirements.js
```diff
@@ -49,6 +65,7 @@ export function populateReqTree(reqTree){
       let reqLabel = <span>
                           <div className='my-arrow'></div>
                           <span className='reqName'>{requirement['name']}</span>
+                          <i className="fa fa-search" onClick={(e)=>{getReqCourses(requirement['path_to'])}}></i>
```
bnehoran, line 70: Add some spacing between the requirement name and the magnifying glass icon. 

Should look maybe something like this:
image

frontend/src/Requirements.js
```diff
@@ -33,6 +34,21 @@ function getHash(stringName) {
   return stringName.split("").reduce(function(a,b){a=((a<<5)-a)+b.charCodeAt(0);return a&a},0);
 }
 
+function getReqCourses(req_path){
```
bnehoran, line 38: Maybe a possible temporary solution to the problem of the requirement subtree collapsing when you click on the magnifying glass would be to have the magnifying glass always (force) expand its containing subtree when clicked (whether it is currently collapsed or expanded), overriding the behavior that comes from clicking on the requirement in general.

This behavior (of always expanding) seems like it would be somewhat appropriate, since when you're searching for a requirement, you are focusing in on it, so you probably wouldn't mind it expanding (or you might even want it to do so). It seems much better anyway than having it collapse when clicked (which is what now happens half the time).

I don't know what the complexity of doing this would be, but just a suggestion. I doubt it would be overly difficult.

frontend/src/Requirements.js
bnehoran, line 70: Also, it would be nice if the magnifying glass changed color or something similar on hover to indicate that it's clickable.

bnehoran (changes requested):
Impressive work overall.
I'm really excited that it works!
I know the visual component with the magnifying glasses is just temporary until we get the popovers, so I'm not overly concerned about how it looks. But a couple small adjustments could make it look and feel more natural.

frontend/src/Requirements.js
bnehoran: Or alternatively, if you could make it right-aligned next to the count, that might look okay as well.

bnehoran: I've also noticed that the icon seems a bit big (it's larger than the text font). Would it maybe look better if it were somewhat smaller?

frontend/src/Requirements.js
danielleung27: i think it should still show so the user knows something is happening because some reqs search a lot of courses and dont show up instantly

config/settings/base.py
```diff
@@ -92,7 +92,7 @@
 SECRET_KEY = os.getenv('SECRET_KEY', 'no7bxov1^uh=ksbp-xyw=#%4pn@01naitpdfj=-3*kao-3w93a')
 
 DATABASES = {}
-DATABASES['default'] = dj_database_url.config()
+DATABASES['default'] = dj_database_url.config(default='postgres://tigerpath:e7e5f0ee15f81d80076716add445d5dacc0b74f0e5c27638a3b47e47017d80abb170c5f64b7f385f87e2eb7f0cf6928a6f@67.207.80.210:5432/tigerpath')
```
bnehoran: 😬

frontend/src/Requirements.js
```diff
@@ -15,16 +15,16 @@ export function toggleSettle(course, path_to, settle){
   let courseOnScheduleNotAssigned = '';
   // checks for duplicates on course schedule and grabs courses that do not have the path and courses that do have the path
   allCoursesOnSchedule.each(function(){
-    if($(this).data('reqs').map((path)=>{
-      return path.split('//')[2]
-    }).indexOf(path_to.split('//')[2]) === -1) courseOnScheduleNotAssigned = $(this);
-    else courseOnScheduleAssigned = $(this);
+    if($(this).data('reqs').indexOf(path_to) === -1)
+      courseOnScheduleNotAssigned = $(this);
```
bnehoran: This change introduces bugs in the course settling. (When I add the same course twice and settle it twice, there is still a third copy left over that remains unsettled.)
Only the first 3 elements of the path should be compared. It should be the same as before, but instead compare elements 0 through 2.

bnehoran (changes requested):

frontend/src/Requirements.js
danielleung27: I thought you said to compare the whole string now? So I should change it back to what it was before?

frontend/src/Requirements.js
```diff
@@ -15,8 +15,9 @@ export function toggleSettle(course, path_to, settle){
   let courseOnScheduleNotAssigned = '';
   // checks for duplicates on course schedule and grabs courses that do not have the path and courses that do have the path
   allCoursesOnSchedule.each(function(){
-    if($(this).data('reqs').indexOf(path_to) === -1)
-      courseOnScheduleNotAssigned = $(this);
+    if($(this).data('reqs').map((path)=>{
+      return '//'.join(path.split('//')[:3])
```
bnehoran: This gives me a compile error:
image

bnehoran (approved):
Looks good

PR #121, opened 2018-07-22
Add popovers for requirements

danielleung27:
Added popovers for requirements and moved search by requirement button to popovers

### DIFF
--- a/frontend/src/Search.js
+++ b/frontend/src/Search.js
@@ -99,7 +99,7 @@ function renderRequirements(){
               if((mainReq['min_needed'] === 0 && mainReq['count'] >= 0) || 
                 (mainReq['min_needed'] > 0 && mainReq['count'] >= mainReq['min_needed']))
                   finished='req-done';
-              let popoverContent = '<div className="popoverContentContainer">';
+              let popoverContent = '<div class="popoverContentContainer">';
               if(mainReq.explanation) {
                 popoverContent += '<p>' + mainReq.explanation + '</p>';
               }
--- a/frontend/src/Search.js
+++ b/frontend/src/Search.js
@@ -127,7 +127,6 @@ function renderRequirements(){
               }
               popoverContent += '</div>'
               let mainReqLabel = <div className='reqLabel' 
-                                  reqpath={mainReq['path_to']}
                                   title={'<span>' + mainReq.name + '</span>'}
                                   data-content={popoverContent}>
                                     <div className='my-arrow root-arrow'></div>
--- a/frontend/src/Search.js
+++ b/frontend/src/Search.js
@@ -11,6 +11,7 @@ import TreeView from 'react-treeview/lib/react-treeview.js';
 import {toggleSettle} from './Requirements';
 import {populateReqTree} from './Requirements';
 import {makeNodesClickable} from './Requirements';
+import {addReqPopovers} from './Requirements';
 
 var dragula = require('react-dragula');
 var current_request = null;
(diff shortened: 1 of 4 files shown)


bnehoran:
I've played around with it a bit, and it looks like it's going in the right direction, but stills some kinks to work out.
I still need to go through the code more thoroughly, since I haven't had a chance to do that yet, but just from playing with it, I've noticed that the popovers for the courses now sometimes appear in the wrong place. Do you know why that happens?

Here's an example: 
image

It seems like it's always for courses in the rightmost column.

bnehoran:
It's also somewhat difficult to "catch" the req popovers, since when I start moving my mouse in the direction of the popover, the moment I move off of the box in the req tree, the popover I was trying to catch disappears. The only way is to move really fast to get there before it disappears.

danielleung27:
The popup appearing in the wrong place should be an easy fix. Its because of my styling on left popovers that I did to fix the req popovers. For the issue about it being hard to catch, would it help if the popover was closer to the req tree? I can move it closer if it helps

churichard (changes requested):
Haven't looked at the code yet, but a couple of points:

- It looks like you're using the explanation from one level higher than you should be.
- Long explanation text in a requirement near the top of the page gets cut off the page.
- You're showing the contacts and urls when you hover over the major/degree. Is that something we want to do? The urls in particular look a bit not user friendly.

danielleung27:
Yeah I can't really figure out how to prevent the long text from getting cut from the top. Popover positioning is really hard, the popover seems to be placed above or below the arrow at random points, I also had to use a css to offset the popovers to the left, its default placement covers the requirements. For the contacts and urls, where else would we put them?

bnehoran (changes requested):
- As Richard said, the explanations are being pulled from the parent of the requirement rather than the requirement itself. (Note that by contrast, the titles _are_ coming from the correct requirement. Only the explanation isn't.) That definitely needs to be fixed before approval. (**critical**)

- The popover for the major should include the major's explanation field, if it exists (in addition to the currently included info).

- The `\n` character should be treated appropriately as a newline to get the proper formatting for the explanations. Right now, they're all clumped up into a single paragraph, which is really hard to read.

- The text inside the search button "Search Requirement" doesn't seem clear enough about what it actually does. (Does it search Google? the registrar's office? What kind of search?) Maybe something more like "Search for courses" or "Find courses" or "Search by this requirement", or something else that makes it more clear. Not sure what the best help text would, but something similar that's clear what it does as well as concise.

- I actually like that the contacts are shown, though I'd like the emails to be made into email links (though not super important). In regards to the links, I'm not sure. Maybe at the very least, they should have a heading "Relevant links", and each link's text should be cut off with an ellipsis (...) to fit on one line. Probably not as important as the other issues, though.

churichard:
Perhaps "Find satisfying courses" would be good for the button

bnehoran:
Apparently, long explanations near the bottom also get cut off:

image

I think it's just going in the wrong direction in these instances. If it's in the lower part of the screen, it should be going up, and if it's in the upper part of the screen, it should be going down.

The default should probably be set to going down, since it looks nicer, and since being cut off at the bottom seems to be less egregious.

bnehoran:
Wow, that looks much better!

Two remaining issues for me:
- The popovers are still really difficult to catch. I think they've moved even farther away than they were before (!) Should be an easy fix, right? 
- The overflow outside the screen is still annoying. If we can't fix it and it comes down to just that, I think it's not the end of the world (we can issue a fix later), but it would be great if this could be fixed.

In my opinion, the other things, we could live with and come back to them after launch.

bnehoran:
Popover placement is inconsistent across different size screens:
image
image

frontend/src/Search.js
```diff
@@ -115,6 +118,7 @@ function renderRequirements(){
               }
               popoverContent += '</div>'
               let mainReqLabel = <div className='reqLabel' 
+                                  reqpath={mainReq['path_to']}
```
churichard: What's this used for?

bnehoran: It's not used.

bnehoran (approved):

churichard (approved):

Issue #120, opened 2018-07-16
Color the words purple, blue, and orange in the tutorial.


bnehoran:
Color the words **purple**, **blue**, and **orange** in the tutorial.

![image](https://user-images.githubusercontent.com/9757500/42788446-8fe14db4-892d-11e8-8c2e-25a28e064e86.png)

Should be an easy thing to do.

churichard:
Merged.

PR #123, opened 2018-08-26
Add colors to tutorial text

bnehoran:
Give color to the color words in the tutorial.

It looks like this:
image

Resolves: #120


bnehoran:
Addresses #120

churichard (approved):
Looks good to me. The colors don't exactly match the search result colors, but I think this is better than the alternatives (which would be using the washed-out background colors or using the much too dark text colors).

Issue #128, opened 2018-08-26
Add ORF major


bnehoran:
A possible start would be 
https://github.com/PrincetonUniversity/majors-and-certificates/blob/master/majors/xxORF_2018.json

Much of the work seems to be completed here, but it's just not formatted correctly, and needs to be double-checked against the official department requirements.

Issue #100, opened 2018-05-16
Course List macros should have the ability to exclude certain courses


bnehoran:
for example, to represent:
All 300 level COS courses except COS 397 and COS 398 (which are independent work)

Not sure how that should be written in the list.
Maybe something like
```javascript
"course_list": [
  "COS 3** -COS 397 -COS 398: Any 300-level COS course except IW",
  "COS 4** -COS 497 -COS 498: Any 400-level COS course except IW"
]
```

With the comments (everything after the colon is ignored anyway), it would be
```javascript
"course_list": [
  "COS 3** -COS 397 -COS 398",
  "COS 4** -COS 497 -COS 498"
]
```

This doesn't look that great though, so maybe someone has a better idea for how to do this.

bnehoran:
or maybe this looks better?
```js
"course_list": [
  "-COS 397",
  "-COS 398",
  "-COS 497",
  "-COS 498",
  "COS 3**",
  "COS 4**"
]
```

bnehoran:
What's better? Adding it directly into the course_list as above, or putting it in a separate json field?
```JavaScript
"excluded_courses": [
  "COS 397",
  "COS 398",
  "COS 497",
  "COS 498"
],
"course_list": [
  "COS 3**",
  "COS 4**"
]
```

This might eliminate the problem of needing to read the negative items in the course_list before the positive ones, although it's not really that big a problem.

This might also have the extra advantage that it's backwards compatible. If the verifier doesn't recognize negative items, then it will still work as before. Though I don't know if backwards compatibility is even useful here.

bnehoran:
We could also potentially decide to support more than one way of notating it, which would not be hard to do, but might be clunky and confusing.

churichard:
I like adding it directly into the course_list. That way, you can't "forget" to exclude certain courses from counting.

bnehoran:
I'm not sure what you mean by forget? 
Do you mean, if we add it to the JSONs but forget to add that functionality to the verifier? Not sure why that would happen by mistake. It could be that we do that because we want to make changes in smaller steps, though.

Also, how would the information then be passed up in the collapsed course lists?
Courses could be excluded from one subtree, but not from another, and so just merging them wouldn't work.

PR #134, opened 2018-08-26
Add ORF major

danielleung27:
Resolves: #128

### DIFF
--- a/tigerpath/majors_and_certificates/majors/ORF_2018.json
+++ b/tigerpath/majors_and_certificates/majors/ORF_2018.json
@@ -2,12 +2,13 @@
     "type": "Major",
     "name": "Operations Research and Financial Engineering",
     "code": "ORF",
+    "degree": "BSE",
     "year": 2018,
     "description": "Operations research and financial engineering (link is external)may be considered as the modern form of a liberal education: modern because it is based on science, mathematics, computing and technology, and liberal in the sense that it provides for broad intellectual development and can lead to many different types of careers. By choosing judiciously from courses in engineering, science, mathematics, economics, public policy, and liberal arts, each student may design a program adapted to his or her particular interests.",
     "urls": [
-        "https://ua.princeton.edu/academic-units/department-operations-research-and-financial-engineering#"
+        "https://ua.princeton.edu/academic-units/department-operations-research-and-financial-engineering#",
+        "https://orfe.princeton.edu/"
     ],
-    "allowed_majors": [],
     "contacts": [
         {
             "type": "Departmental Representative",
@@ -23,12 +24,11 @@
     "req_list": [
         {
             "name": "Core",
-            "max_counted": 4,
+            "max_counted": 1,
             "min_needed": 4,
             "explanation": "The core requirements (four courses). These form the intellectual foundation of the field and cover statistics, probability, stochastic processes, and optimization, along with more advanced courses in mathematical modeling.",
             "double_counting_allowed": false,
             "completed_by_semester": 8,
-            "req_list": [],
             "course_list": [
                 "ORF 245",
                 "ORF 307",
@@ -38,7 +38,7 @@
         },
         {
             "name": "Departmental Electives",
-            "max_counted": 10,
+            "max_counted": 1,
             "min_needed": 10,
             "explanation": "These are courses that either extend and broaden the core, or expose the student to a significant problem area or application closely related to the core program. \n1. There must be at least four courses from the Department of Operations Research and Financial Engineering (ORF).\n2. There can be no more than three courses from any one department (excluding ORF).",
             "double_counting_allowed": false,
@@ -239,19 +239,18 @@
                         "NEU 330"
                     ]
                 }
-            ],
-            "course_list": []
+            ]
         },
         {
             "name": "Senior Independent Research",
             "max_counted": 1,
-            "min_needed": 1,
+            "min_needed": 2,
             "explanation": "A full-year thesis (or a one semester project plus an additional 400-level ORFE departmental) involving an application of the techniques in the program applied to a topic that the student chooses in consultation with a faculty adviser.",
             "double_counting_allowed": false,
             "completed_by_semester": 8,
-            "req_list": [],
             "course_list": [
-                "ORF 478"
+                "ORF 478",
+                "ORF 479"
             ]
         }
     ]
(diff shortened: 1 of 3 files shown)


bnehoran:
Addresses #128

tigerpath/majors_and_certificates/majors/ORF_2018.json
```diff
@@ -0,0 +1,258 @@
+{
+    "type": "Major",
+    "name": "Operations Research and Financial Engineering",
+    "code": "ORF",
+    "year": 2018,
```
bnehoran: Needs "degree" field

tigerpath/majors_and_certificates/majors/ORF_2018.json
```diff
@@ -0,0 +1,258 @@
+{
+    "type": "Major",
+    "name": "Operations Research and Financial Engineering",
+    "code": "ORF",
+    "year": 2018,
+    "description": "Operations research and financial engineering (link is external)may be considered as the modern form of a liberal education: modern because it is based on science, mathematics, computing and technology, and liberal in the sense that it provides for broad intellectual development and can lead to many different types of careers. By choosing judiciously from courses in engineering, science, mathematics, economics, public policy, and liberal arts, each student may design a program adapted to his or her particular interests.",
+    "urls": [
+        "https://ua.princeton.edu/academic-units/department-operations-research-and-financial-engineering#"
+    ],
+    "allowed_majors": [],
```
bnehoran: Remove
This is only for certificates.

tigerpath/majors_and_certificates/majors/ORF_2018.json
```diff
@@ -0,0 +1,258 @@
...
+        }
+    ],
+    "req_list": [
+        {
+            "name": "Core",
+            "max_counted": 4,
+            "min_needed": 4,
+            "explanation": "The core requirements (four courses). These form the intellectual foundation of the field and cover statistics, probability, stochastic processes, and optimization, along with more advanced courses in mathematical modeling.",
+            "double_counting_allowed": false,
+            "completed_by_semester": 8,
```
bnehoran: `double_counting_allowed` is false by default, and `completed_by_semester` is 8 by default, so no need to include explicitly. They don't hurt necessarily, but I've been adding them only when they're not the default values, so that it's less cluttered.
We might want to decide on a standard (whether to be explicit or implicit with these).

tigerpath/majors_and_certificates/majors/ORF_2018.json
```diff
@@ -0,0 +1,258 @@
...
+            "req_list": [],
```
bnehoran: Remove this. The interpreter assumes only a `req_list` *or* a `course_list` exists, so it does not get interpreted properly if they both appear.
Yes, the form spits them both out (sorry about that, but I couldn't find a way around it), so unfortunately, the empty one just has to be removed afterwords (easy with find and replace).

tigerpath/majors_and_certificates/majors/ORF_2018.json
```diff
@@ -0,0 +1,258 @@
...
+                {
+                    "name": "ORF",
+                    "max_counted": 10,
+                    "min_needed": 4,
+                    "explanation": "1. There must be at least four courses from the Department of Operations Research and Financial Engineering (ORF).\n2. There can be no more than three courses from any one department (excluding ORF).",
+                    "double_counting_allowed": false,
+                    "completed_by_semester": 8,
+                    "req_list": [],
+                    "course_list": [
+                        "ORF 311",
```
bnehoran: Are these all the upper level courses in the ORF department?
If it is, they can be specified by 
`ORF 3**` and `ORF 4**`
It's easier to read/debug and more maintainable in case the courses available change.
If not (or if some courses are left out) then this is okay for now. In the future, we should add an ability to exclude courses as mentioned in #100

tigerpath/majors_and_certificates/majors/ORF_2018.json
```diff
@@ -0,0 +1,258 @@
...
+            "name": "Departmental Electives",
+            "max_counted": 10,
+            "min_needed": 10,
+            "explanation": "These are courses that either extend and broaden the core, or expose the student to a significant problem area or application closely related to the core program. \n1. There must be at least four courses from the Department of Operations Research and Financial Engineering (ORF).\n2. There can be no more than three courses from any one department (excluding ORF).",
+            "double_counting_allowed": false,
+            "completed_by_semester": 8,
+            "req_list": [
+                {
+                    "name": "ORF",
+                    "max_counted": 10,
```
bnehoran: No need for this `max_counted` since it's enforced by the parent, right?

tigerpath/majors_and_certificates/majors/ORF_2018.json
bnehoran: Also in multiple other places.

tigerpath/majors_and_certificates/majors/ORF_2018.json
bnehoran: Also appears in multiple other places

tigerpath/majors_and_certificates/majors/ORF_2018.json
```diff
@@ -0,0 +1,258 @@
...
+                    ]
+                }
+            ],
+            "course_list": []
+        },
+        {
+            "name": "Senior Independent Research",
+            "max_counted": 1,
+            "min_needed": 1,
+            "explanation": "A full-year thesis (or a one semester project plus an additional 400-level ORFE departmental) involving an application of the techniques in the program applied to a topic that the student chooses in consultation with a faculty adviser.",
```
bnehoran: From the description, it seems like there are two options for this requirement?

It looks like only only one is listed.

bnehoran (changes requested):
* Some extra fields (empty req and course lists, `completed_by_semester`, `allowed_majors`, `double_counting_allowed`) should be removed to conform to schema.
* Should to be tested with frontend to ensure proper functionally (not clear if this has been done)
* At least one test should be written in `verifier_tests` (to ensure that future changes in verifier don't break this major's functionally)
* I haven't reviewed the content very thoroughly against the departmental page, so that needs to be done as well.

tigerpath/majors_and_certificates/majors/ORF_2018.json
danielleung27: There was a list provided, it might not be all the upper level departmentals for orf.

bnehoran: If it's from an explicit list, then that's fine. Best to have it the same way they specify it.

bnehoran (approved):

bnehoran:
Still need to add test cases in `verifier_tests`, but this can be done later.

Issue #129, opened 2018-08-26
Add CEE major


bnehoran:

churichard:
Merged, so I'm closing this.

PR #135, opened 2018-08-26
Add CEE major

danielleung27:
Add CEE major

Resolves: #129


bnehoran:
Addresses #129

tigerpath/majors_and_certificates/majors/CEE_2018.json.txt
```diff
@@ -0,0 +1,846 @@
...
+    "urls": [
+        "https://ua.princeton.edu/academic-units/department-civil-and-environmental-engineering#"
+    ],
+    "allowed_majors": [],
+    "contacts": [
+        {
+            "type": "Departmental Representative",
+            "name": "Branko Gli�ic",
```
bnehoran: Do we support characters with diacritics? How well do they display in the frontend?

tigerpath/majors_and_certificates/majors/CEE_2018.json.txt
```diff
@@ -0,0 +1,846 @@
+{
```
bnehoran: Should have a .json extension.

bnehoran (changes requested):
See comments for similar issues in #134

bnehoran:
Addressed some of the format issues so that it can be viewed/used inside the app to more easily test it and check the content.

Still needs to be tested and reviewed for content/correctness as mentioned in https://github.com/TigerPathApp/tigerpath/pull/134#pullrequestreview-151609789

bnehoran (changes requested):
Still need to allow double-counting across tracks.

Also, should consider factoring out requirements that are common across all tracks.
This would make it easier for users to navigate, as well as for us to maintain.

danielleung27:
Why should double counting be allowed across tracks? Students can only do 1 track I think? Also factoring out the requirements might be a little confusing. The CEE website separates the 5 tracks with requirement lists without factoring out common courses so a user familiar with how it is structured on the website will get confused here.

bnehoran (approved):
I've gone through the content and compared it to the official requirements pages.
I also did a bit of testing within the app, but it's difficult to do at the moment because there are so few semesters of courses, that many of the courses required for the major don't exist. So more testing should be done at some point to check that all the requirements can indeed be fulfilled within the app.

I fixed any problems I found as I discovered them. Most seem to be due to confusion from the inconsistencies between different sources (for instance, additional math and science requirements only appear in the CEE "Yello Book").
The requirements for the "Engineering and the Liberal Arts" track still look suspect, but it's unclear from the undergraduate announcement or from the department web page what the requirements actually are.
In both cases, we need to verify with someone who is more familiar with the major.

Lastly, still no test cases in `verifier_tests`, but I think this is okay for now. We can add some later.

PR #136, opened 2018-08-27
Add logging in without using CAS

churichard:
Added ability to log in using Django's authentication system directly without using CAS. This is only enabled in development mode, and is so that we can log in even after our Princeton NetIDs don't work anymore.

Please check that this doesn't break the regular logging in using Princeton CAS (I can't check this).


danielleung27 (approved):
CAS login still works normally for me. Looks good.

PR #144, opened 2018-09-03
Add SSL support

churichard:
Upgraded Heroku to a hobby dyno, enabled HTTPS on their end, and added SSL support in Django. The dev site needs to be tested after merging, can't test locally.

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -14,12 +14,11 @@ WEBPACK_LOADER = {
         }
 }
 
+
 # Security
-# CSRF_COOKIE_SECURE = True  # should be uncommented when SSL is implemented
-
-# SESSION_COOKIE_SECURE = True  # should be uncommented when SSL is implemented
-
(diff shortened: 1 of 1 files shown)


danielleung27 (approved):
Not really sure what this does but looks good

churichard:
**CSRF_COOKIE_SECURE**: Whether to use a secure cookie for the CSRF cookie. If this is set to True, the cookie will be marked as “secure,” which means browsers may ensure that the cookie is only sent with an HTTPS connection.

**SESSION_COOKIE_SECURE**: Whether to use a secure cookie for the session cookie. If this is set to True, the cookie will be marked as “secure,” which means browsers may ensure that the cookie is only sent under an HTTPS connection.

**SECURE_SSL_REDIRECT**: If True, the SecurityMiddleware redirects all non-HTTPS requests to HTTPS (except for those URLs matching a regular expression listed in SECURE_REDIRECT_EXEMPT).

Source: https://docs.djangoproject.com/en/2.1/ref/settings/

Issue #125, opened 2018-08-26
Replace search text when searching by requirement


bnehoran:
When searching by requirement, the existing search text in the search box should either be cleared and be blank , or, even better, be replaced with a the title of the requirement being searched by.

For example, suppose the a COS student searches for COS126
![image](https://user-images.githubusercontent.com/9757500/44629956-a95a8a00-a924-11e8-91f8-df23b963eaf3.png)

Then they search by requirement for COS->Core->Theory, it currently looks like this
![image](https://user-images.githubusercontent.com/9757500/44629970-ee7ebc00-a924-11e8-84d7-2f7a91ccac68.png)

This ends up being confusing for many users.

Instead it should display like this (or something similar):
![image](https://user-images.githubusercontent.com/9757500/44629944-7dd79f80-a924-11e8-9f40-aec209688394.png)

PR #146, opened 2018-09-04
Search text changed on search by req

adeniji19:
I think the best choice is to put generic text so the user knows what is being displayed but doesn't think they can replicate by typing it into the search bar. 

Resolves: #125

### DIFF
--- a/frontend/src/Requirements.js
+++ b/frontend/src/Requirements.js
@@ -36,6 +36,7 @@ function getHash(stringName) {
 }
 
 function getReqCourses(req_path){
+  document.getElementsByName('search-text')[0].value='Filtered Search';
   $('#spinner').css('display', 'inline-block');
   $.ajax({
       // the slashes messes up the url 
--- a/frontend/src/Requirements.js
+++ b/frontend/src/Requirements.js
@@ -36,7 +36,8 @@ function getHash(stringName) {
 }
 
 function getReqCourses(req_path){
-  document.getElementsByName('search-text')[0].value='Filtered Search';
+  var filterText = req_path.substring(req_path.lastIndexOf("//")+2,req_path.length) + ' Requirement';
+  document.getElementsByName('search-text')[0].value=filterText;
   $('#spinner').css('display', 'inline-block');
   $.ajax({
       // the slashes messes up the url 
--- a/frontend/src/Requirements.js
+++ b/frontend/src/Requirements.js
@@ -36,7 +36,7 @@ function getHash(stringName) {
 }
 
 function getReqCourses(req_path){
-  var filterText = req_path.substring(req_path.lastIndexOf("//")+2,req_path.length) + ' Requirement';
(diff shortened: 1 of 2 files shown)


churichard (changes requested):
I think displaying the requirement title would be more informative, since the user might forget what was searched.

bnehoran:
I removed the commits that were unrelated to this pull request, so if your local branch still has those commits but you want to pull, you just have to do `git reset --hard master` before doing `git pull`.

----------------------------
Caveat: If you have local, unpushed commits you want to keep, you might want to first make a copy of the branch. For instance,
```bash
git branch old_search_text
git reset --hard master
git pull
git cherry-pick old_search_text~[n]..old_search_text #replace the [n] with the number of commits to keep
```

bnehoran:
I removed the commits that were unrelated to this pull request and force pushed, so if your local branch still has those commits but you want to pull, you just have to do `git reset --hard master` before doing `git pull`.

----------------------------
Caveat: If you have local, unpushed commits you want to keep, you might want to first make a copy of the local branch. For instance,
```bash
git branch old_search_text
git reset --hard master
git pull
git cherry-pick old_search_text~[n]..old_search_text #replace the [n] with the number of commits to keep
```

bnehoran (approved):

bnehoran:
Resolves: #125

frontend/src/Requirements.js
```diff
@@ -36,6 +36,8 @@ function getHash(stringName) {
 }
 
 function getReqCourses(req_path){
+  var filterText = 'Satisfying: ' + req_path.substring(req_path.lastIndexOf("//")+2,req_path.length);
```
churichard: A cleaner way to do this is by doing `req_path.split('//').pop()`

frontend/src/Search.js
```diff
@@ -309,6 +309,7 @@ class Search extends Component {
     return (
         <div>
         <input type = "text" 
+          name="search-text"
```
churichard: Don't use `name`, that's only used in forms. If you want to identify an element, use the field `id`

frontend/src/Requirements.js
```diff
@@ -36,6 +36,8 @@ function getHash(stringName) {
...
+  document.getElementsByName('search-text')[0].value=filterText;
```
churichard: jQuery can do this more cleanly: `$('#search-text').val(filterText);`

churichard (changes requested):

frontend/src/Search.js
```diff
@@ -305,10 +305,11 @@ class Search extends Component {
 
   }
   render()
-  { 
+  {
     return (
         <div>
-        <input type = "text" 
+        <input type = "text"
+          input id="search-text"
```
churichard: Don't need `input` here, just `id` is fine.

churichard (changes requested):

churichard (approved):

PR #147, opened 2018-09-06
Update ELE_2018.json

adeniji19:
correcting specific ele 208b requirement.

### DIFF
--- a/tigerpath/majors_and_certificates/majors/ELE_2018.json
+++ b/tigerpath/majors_and_certificates/majors/ELE_2018.json
@@ -41,7 +41,7 @@
       "course_list": [
         "ELE 301",
         "ELE 302",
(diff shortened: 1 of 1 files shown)


bnehoran:
Could you provide a source link (as in, where this requirement is explained)?
Thanks!

bnehoran (approved):
I'm approving it, but I'm still wondering if you can attach a link to an official ELE source that explains the ELE requirements in regards to this course. Perhaps provide the link as a comment before you merge, so that it's in the records if we need to look back at it in the future.

adeniji19:
http://ee.princeton.edu/sites/default/files/UG_Handbook.pdf

churichard:
This was merged into master instead of dev. I removed it from master and added it to dev instead.

PR #148, opened 2018-09-06
Fix scraper

danielleung27:
Prevents bug when rescraping a semester, fixes bug where semesters list would be blank if you scraped a course again for the same semester.


churichard (approved):
Looks fine, but the ELE change shouldn't be in here. Can you remove that?

PR #174, opened 2018-12-22
Refactor React components, use react-beautiful-dnd for drag and drop

churichard:
Been working on this change for a few weeks now, it's a big one!

I'm refactoring all of our React code into more modular components so that it's more readable. It also conforms more to best practices for how you're supposed to write React code. There's still room for improvement at the low-level of how we do things, but hopefully this PR makes it so that it's easier to maintain this code in the future. A lot of the code we had written before was very tightly coupled together, which is why this refactoring PR had to be so big -- one refactor just kept leading to another 😅

As a result of this change, I've had to drop `dragula` as our drag-and-drop library of choice, because it doesn't actually play very well with how React works. If you're interested in the specifics, `dragula` changes the DOM directly when you drag-and-drop, but in React you're supposed to change the state and leave the DOM updating up to React. Instead of using `dragula`, I've opted for [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd), which is actually a pretty nice library. There are a couple of additional small issues that using this library creates, which I hope to address in a future PR (this one is big enough as it is).

The issues I've identified are:
- dragging and dropping from search results into a semester creates an empty space that is the same height as the search result course card
- when you start dragging a course, there is no indication of which semester you will actually be putting the course into once you let go

The behavior of everything else should basically be identical to how it was before.


frontend/src/components/App.js
```diff
@@ -0,0 +1,145 @@
...
+          data = data.map(mainReq => {
+            // if mainReq is not an array, then it is just the name of the major
+            if (!Array.isArray(mainReq)) return mainReq
+            return mainReq[2];
+          });
+          this.setState({requirements: data});
+        }
+      }
+    };
+    if (schedule) {
```
danielleung27, line 91: maybe add a comment about what's happening here? i wasn't really sure

danielleung27:
Other minor UI bugs:
- if you drag a course from 1 semester to another, the space from the source of the drag doesn't close up until you drop the course somewhere else, and it closes up the space instantly and not fluidly like the rest of the drag and drops.
- dragging a course makes the shadow for the dragged course disappear

I think it might be fine to leave out the ghost image indicator for where the course will be dropped. The drag and drop UI is more fluid with react-beautiful-dnd so it might be a little awkward to show the ghost of the course instantly since it takes some time for the courses to be pushed away now.

danielleung27 (approved):

churichard:
Thanks for looking at this PR!

The first bug you mentioned is just how `react-beautiful-dnd` does it for some reason - if you look at the examples on their GitHub repo, they have the same issue. The second bug is definitely worth fixing though.

When you talk about the ghost image indicator, are you referencing the second bug that I identified in my original comment ("when you start dragging a course, there is no indication of which semester you will actually be putting the course into once you let go")? If so, I think this is still a problem because when there are no courses in the semester, you don't actually know what'll happen when you let go. I think a good way to fix this might be to highlight the semester that you're currently dragging over, so you know that the course will go into that semester. This has the added benefit of also allowing us to have a different highlight color depending on whether the course fits in that semester or not (e.g. the highlight color could be red if you are trying to put a fall course into a spring semester).

danielleung27:
Oh ok, I guess that's fine then. Running locally makes everything lag a bit for me so it looked a bit awkward because the gap from the source of the drag didn't close up immediately when the drop occurred, it might be faster when deployed.

I think highlighting which semester the drop is occurring on is a good idea.

PR #194, opened 2019-01-31
Open jsons with utf8

danielleung27:
Opening with just 'r' will crash on jsons with unicode characters. Opening with utf8 solves this.

### DIFF
--- a/tigerpath/majors_and_certificates/scripts/verifier.py
+++ b/tigerpath/majors_and_certificates/scripts/verifier.py
@@ -173,7 +173,7 @@ def get_courses_by_path(path):
         req_filepath = os.path.join(_get_dir_path(), DEGREES_LOCATION, filename)
     else:
         raise ValueError("Path malformatted.")
-    with open(req_filepath, encoding="utf8") as f:
+    with open(req_filepath, 'r', encoding="utf8") as f:
         req = json.load(f)
     subreq = _get_req_by_path(req, path)
     if not subreq:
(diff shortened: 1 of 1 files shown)


churichard (approved):

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -120,7 +120,7 @@ def check_requirements(req_file, courses):
     :returns: A simplified json with info about how much of each requirement is satisfied
     :rtype: (bool, dict, dict)
     """
-    with open(req_file, 'r') as f:
+    with open(req_file, encoding="utf8") as f:
```
bnehoran: The encoding shouldn't be instead of the mode, but in addition.
```suggestion
    with open(req_file, 'r', encoding="utf8") as f:
```

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -173,7 +173,7 @@ def get_courses_by_path(path):
         req_filepath = os.path.join(_get_dir_path(), DEGREES_LOCATION, filename)
     else:
         raise ValueError("Path malformatted.")
-    with open(req_filepath, 'r') as f:
+    with open(req_filepath, encoding="utf8") as f:
```
bnehoran:
```suggestion
    with open(req_filepath, 'r', encoding="utf8") as f:
```

bnehoran (changes requested):

bnehoran (approved):

Issue #173, opened 2018-12-19
Proposal: Use YAML instead of JSON for encoding requirement specifications


bnehoran:
YAML allows for more flexibility than JSON, and might allow us to factor out requirements to make them more readable and maintainable.
https://en.wikipedia.org/wiki/YAML

In particular, YAML has references and anchors, which we could use, for example, to encode the course lists separately from the structure of the major, and then just link to those course lists from within the structure.
Then, whenever the course lists change, we don't have to search through the structure to find each of the places where the course lists appears in order to update it. This would make it easier to encode and to maintain.
YAML also allows comments and other features that would be very nice to have for maintainability.

JSON is a subset of YAML, so the existing requirement JSON files wouldn't initially need to be changed and should work as is.
New requirements can be written in a simpler way, and we can update the old ones gradually to simplify them.

Maybe there are other formats that are just as good or better. I haven't really done an extensive search, but YAML seems to have this advantage of compatibility with the existing JSONs and at the same time adding useful functionality.

churichard:
I think that's an interesting idea. It definitely has the potential to make it easier to create and maintain the requirement specs.

I think the one downside with YAML is that it might not have the first-class support that JSON has, so it may be a little more annoying to parse through. The benefits of using it might outweigh this downside, though.

Could you try encoding an existing major in YAML to see what it would look like (e.g. COS BSE)?

bnehoran:
I've switched to a YAML parser, and everything works without a problem with no other changes needed.

bnehoran:
This allows writing neater requirements.
For example, this is what the COS-BSE major looks like as a (simplified) YAML:
https://github.com/TigerPathApp/tigerpath/blob/verifier-updates/tigerpath/majors_and_certificates/majors/COS-BSE_2018.yaml

Compare to the original JSON (/unsimplified YAML):
https://github.com/TigerPathApp/tigerpath/blob/verifier-updates/tigerpath/majors_and_certificates/majors/COS-BSE_2018.json

Of course, we don't need to convert any of the existing JSONs if we don't want to, but any new requirements can made cleaner.

Issue #101, opened 2018-05-16
Add support for requirements to be satisfied by AP credits, summer courses, study abroad, etc.


bnehoran:
Add support for requirements to be satisfied by AP credits, as well as courses taken during the summer or during study abroad.

Similarly, allow users to mark that they have received a waiver for a requirement.

bnehoran:
Additional question: If we allow force completion of a requirement, would we allow partial force completion?
For example, if you need 4 of a particular requirement, then could a user say that they already have 1, and then the app only requires 3 more using courses in their schedule?

churichard:
Yes, I think we should allow partial force completion. For example, for COS, I only satisfied one systems course when I studied abroad, but I need two systems courses for that subrequirement.

bnehoran:
**Here's the proposal:**
We add a 9th semester to the user course schedule, which will now contain the pseudo-courses that are manually added by the user with custom names.
These will always be added as "settled" courses, with the `settled` field being populated with the path to the requirement they satisfy.
If done right, the pseudo-course will automatically be added by the verifier to the right place in the requirements tree, and will appear as if it were just another course, but with the user's custom name.

Here are the steps we need to take:

- [ ] Add another pseudo-semester to the user schedule in the database, and ensure that the app only ever reads off the 8 real semesters everywhere else in the app, including the GUI. Check that this doesn't break any functionality. 
- [ ] Pass the now 9-semester schedule to the verifier as input and check that this doesn't break anything, that it still works as intended. I think the only modification needed here to get this to work is to remove any sanity checks for no more than 8 semesters, if these exist in the code.
- [ ] On the GUI side, add a button to each req popover that says something like "Add external or AP credit". When pressed, this should bring up a popup with a text field allowing the user to enter a name for the class, and an explanation/warning that this course will be added only to this req, and without any verification that it actually satisfies the req.
- [ ] When a user fills in this text field and submits, the app should add a pseudo-course, with the user-defined title and the path to the req from which it originated as its `settled` field, to the user's extra pseudo-semester, and then send a new request to refresh the req tree.


Two additional things we need to consider:
- Should the extra semester be at the beginning of the list, in position 0, or at the end, in position 8? Which one makes more sense, and which one would make the conventions simpler and modifications easier?
[...]

PR #202, opened 2019-02-02
Interpret reqs as YAML, enable external credits and exclusion in verifier

bnehoran:
- Requirement JSONs are now interpreted as YAML files.
As expected, this passes all tests without any changes needed.
- A few tests are updated to conform to recent changes in the COS and MAE major requirements (from earlier merged pull requests; we just forgot to update the tests back then)
- Minor changes to the verifier to allow counting external credits (AP credits, summer courses, study abroad, transfer credits)
- Enable excluding courses from a req by adding an `excluded_course_list` field

Addresses: #101 (still need frontend support to fully resolve)
Resolves: #173 
Resolves: #100


tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -530,43 +546,51 @@ def _course_match(course_name, pattern):
     pattern = pattern.split(':')[0] # remove course title
     pattern = ["".join(p.split()).upper() for p in pattern.split('/')] # split by '/' and
     course = ["".join(c.split()).upper() for c in course_name.split('/')] # remove spaces
+    match = False
     for c in course:
         for p in pattern:
+            if p[0] == '~':  # excluded course
```
churichard: Remove support for ~

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -392,7 +396,8 @@ def _init_path_to(req):
             _init_path_to(subreq)
 
 def _json_format(obj):
-   return json.dumps(obj, sort_keys=False, indent=2, separators=(',', ': ')) + "\n"
+    import json
```
churichard, line 399: Is there a reason you moved the import here? Might be better to have it at the top of the file

churichard (changes requested):

tigerpath/majors_and_certificates/scripts/verifier.py
bnehoran: This is just a debugging function and is not called under normal usage, and this is the only place where json is needed.

churichard (approved):
Looks great!

PR #205, opened 2019-02-02
Fix majors

danielleung27:
Small fixes for all the majors


churichard (approved):

churichard:
Might want to update this with the excluded courses once the verifier update is merged in

PR #207, opened 2019-02-02
Update search card padding

churichard:
Use 3px on left and right


danielleung27 (commented):
What do you think of just doing margin: 3px; ? I think it looks more uniform and I feel like 5px is too much.

danielleung27 (approved):

PR #214, opened 2019-02-03
Major fixes

danielleung27:
Double counting should not be allowed for departmentals

Resolves: #215


tigerpath/majors_and_certificates/majors/MAT_2018.json
```diff
@@ -106,7 +106,7 @@
           ]
         },
         {
-          "name": "Geometry or Topology",
+          "name": "Geometry/ Topology or Discrete Math",
```
bnehoran: Doesn't the '/' break the API calls?

tigerpath/majors_and_certificates/majors/MOL_2018.json
```diff
@@ -150,29 +150,37 @@
...
+                  "name": null,
+                  "max_counted": 1,
+                  "min_needed": 1,
+                  "explanation": "",
+                  "course_list": [
+                    "MAT 103",
+                    "MAT 104",
+                    "MAT 175",
+                    "MAT 192",
+                    "MAT 2*"
```
bnehoran, line 181: Do 300 and 400 level classes not count here?
It seems odd that they would restrict it to lower level classes.

tigerpath/majors_and_certificates/majors/POL_2018.json
```diff
@@ -37,7 +37,6 @@
...
-      "double_counting_allowed": true,
```
bnehoran, line 40: Why is double counting removed here?

bnehoran (changes requested):

tigerpath/majors_and_certificates/majors/MAT_2018.json
bnehoran: Maybe change to
```suggestion
          "name": "Geometry, Topology, or Discrete Math",
```

danielleung27: I tried testing it and I think if you put a space it doesn't break it, but yeah maybe we could stick to the commas just in case. But Geometry and Topology is apparently very similar according to feedback and it makes more sense to be Geometry/Topology or Discrete Math. It's hard to do without using special characters

tigerpath/majors_and_certificates/majors/MOL_2018.json
danielleung27: If you look at the explanation in the parent: "...math (MAT 103, 104, 175, 192, or any 200-level MAT course)"

tigerpath/majors_and_certificates/majors/MAT_2018.json
bnehoran: I think it's fine to have commas for now. No one will complain.
On the other hand, I could definitely see '/' causing real problems, and it looks bad that it's not symmetric.

On the longer term, of course, we need to escape our API calls and req paths in particular.

bnehoran (approved):

PR #234, opened 2019-02-07
Update majors

danielleung27:
phy: fix electives and added course for jp and senior thesis

### DIFF
--- a/tigerpath/majors_and_certificates/majors/PHY_2018.json
+++ b/tigerpath/majors_and_certificates/majors/PHY_2018.json
@@ -236,20 +236,34 @@
             "PHY 3*",
             "PHY 4*",
             "PHY 5*",
-            "AST *",
-            "CHM *",
-            "QCB *",
-            "COS *",
-            "GEO *",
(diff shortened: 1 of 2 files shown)


bnehoran (changes requested):
It seems that in order to allow them to add JP and Senior Thesis, we need to allow for courses to not count towards the number of courses taken. I've created an issue for it: #235 
For now, though probably best to leave it unchecked.
So, remove the IW changes for now.

tigerpath/majors_and_certificates/majors/PHY_2018.json
```diff
@@ -236,32 +236,50 @@
...
+            "MSE 5*"
           ]
         }
       ]
     },
     {
       "name": "Independent Work",
-      "max_counted": null,
-      "min_needed": null,
+      "max_counted": 1,
```
bnehoran: Remove these changes

bnehoran (changes requested):

bnehoran (approved):

Issue #179, opened 2019-01-08
Make gray the color for courses that fit in both semesters


bnehoran:
Maybe courses that can fit in either semester should be gray.
It seems more clear visually that they can fit in either semester this way.

This is what it currently looks like:
![image](https://user-images.githubusercontent.com/9757500/50809457-86c47e80-12d1-11e9-93a3-d0e6e662dcfb.png)
Orange courses look like they just belong to a different, third category, which is not the impression we want.

---

And here's an idea of what it would look like:
![image](https://user-images.githubusercontent.com/9757500/50809418-4ebd3b80-12d1-11e9-99f4-de38565db670.png)
I think you can see the dichotomy of colors much better.

churichard:
I agree that you can see which courses match which semesters better in that screenshot, but I'm not sure if grey would be the best color to use. Grey kind of signifies that the course is disabled or inactive in some way, which is not the impression we want to give.

I think having a color that's a mix of the current purple and blue being used would be better. I picked out a middle color using this website: https://meyerweb.com/eric/tools/color-blend/#AD79D4:3AAED4:10:hex and adjusted it a little bit to end up with the following:

<img width="1203" alt="screen shot 2019-01-12 at 9 29 57 pm" src="https://user-images.githubusercontent.com/4218237/51081943-4e49e980-16b1-11e9-9e94-edb65086c413.png">

For reference, the background color is #b2c0e6 and the text color is #1d2848. I think this looks better than the orange currently being used, as well as the proposed grey.

bnehoran:
Yeah, I guess it might make it seem disabled, so I don't know.
The middle color you choose looks to me (personally) much more like the blue than the purple. I don't know if everyone sees it the same way, but I would probably be confused by that.

Maybe what we really need is like a sort of decontrasted striped purple/blue background. Something that clearly indicates "both colors".

And I guess we just really need to add the warning when they add a course in the wrong semester.

PR #236, opened 2019-02-10
Add external credit support

bnehoran:
- Implement external credit support in a separate tab in the main view
- Use styled components instead of a css file to style the schedule + semesters
- Refactor the schedule by using css grid instead of tables, which allows for more flexibility
- Update color for courses that can fit in both semesters to a gradient
- Fix bug where the delete course button was overflowing to the next line in small widths
- Update tutorial colors

Resolves https://github.com/TigerPathApp/tigerpath/issues/101
Resolves https://github.com/TigerPathApp/tigerpath/issues/179

Schedule view:

image

External Credits view:

image


bnehoran:
This looks really good!

The only issue I can see is that when you click on an external credit in the req tree, it disappears from the req tree and there's no way to bring it back other than to delete the class and readd it.

Also, we've discussed this, and maybe it doesn't have to be part of the MVP, but should be added at some point: there should be a way to pin additional requirements on external courses that have already been added. This is both for external credits that satisfy multiple requirements, and so that the user doesn't have to delete and readd them when they switch majors.

frontend/src/components/App.js
```diff
@@ -61,7 +53,12 @@ export default class App extends Component {
     for (let semIndex = 0; semIndex < schedule.length; semIndex++) {
       strippedSchedule.push([]);
       for (let course of schedule[semIndex]) {
-        strippedSchedule[semIndex].push({id: course['id'], settled: course['settled']});
+        let strippedCourse = { id: course['id'], settled: course['settled'] };
```
bnehoran, line 56: How come there's no name for non-external courses?

frontend/src/components/ExternalCreditForm.js
```diff
@@ -0,0 +1,159 @@
...
+                        <Dropdown.Item key={semName} eventKey={index} onSelect={(e) => this.handleChange("selectedSemester", semName, e)}>
+                          {semName}
+                        </Dropdown.Item>
+                      )
+                    }
+                  </Dropdown.Menu>
+                </SemesterDropdown>
+                <Form.Text className="text-muted">
+                  For AP credits, waivers, or other external credits that don't fit in a semester, choose "N/A" - they will appear in "Your External Credits" on the left.
+                </Form.Text>
```
bnehoran:
```suggestion
                  For summer courses, choose either the semester before or after in which you want it to appear.
                </Form.Text>
```

frontend/src/components/ExternalCreditForm.js
```diff
@@ -0,0 +1,159 @@
...
+                      getSemesterNames(profile.classYear).map((semName, index) =>
+                        <Dropdown.Item key={semName} eventKey={index} onSelect={(e) => this.handleChange("selectedSemester", semName, e)}>
+                          {semName}
+                        </Dropdown.Item>
+                      )
+                    }
+                  </Dropdown.Menu>
+                </SemesterDropdown>
+                <Form.Text className="text-muted">
+                  For AP credits, waivers, or other external credits that don't fit in a semester, choose "N/A" - they will appear in "Your External Credits" on the left.
```
bnehoran: What's going on here? Is the apostrophe / single quote supposed to be escaped?

frontend/src/styles/Courses.css
```diff
@@ -42,8 +42,8 @@
 }
 
 .search-card.both {
-  background-color: #fbf2e9;
-  color: #753a00;
+  background: linear-gradient(to right, #f6ebff, #ebf6f9);
```
bnehoran, line 45: This looks a bit odd, but I guess I could get used to it.
It will probably make the color scheme more intuitive, so I think it's an improvement.

tigerpath/static/app/js/tutorial.js
```diff
@@ -11,7 +11,7 @@ function startIntro() {
         intro: `You can search for courses you want to add to your schedule here. This can be by the course code or title.
         Try searching something!
```
bnehoran: It seems that with the newer rewording, I've observed that users no longer take it to mean literally do it now.
This is not really related to this PR, bit it's a small enough change that it probably doesn't merit its own.
```suggestion
         Try searching something now!
```

(and since you're already making changes to this spot in this PR)

tigerpath/views.py
```diff
@@ -242,6 +246,11 @@ def update_schedule(request):
 def get_schedule(request):
     curr_user = request.user.profile
     schedule = populate_user_schedule(curr_user.user_schedule)
+
+    # make sure that the schedule has 9 semesters
+    if len(schedule) < 9:
```
bnehoran:
```suggestion
    while len(schedule) < 9:
```
I don't know if this is needed right now, since all users will presently have either 8 or 9, but it would cover any future changes in schedule length.

bnehoran: Like, suppose we expand to 10 at some point, and some user still has 8.

danielleung27:
I agree with Barak that clicking on a req on the tree from external credit and making it go away doesn't really make sense (unless that's a bug) since the course is still on the schedule. Other than that, everything looks really good!

danielleung27 (dismissed):
I agree with Barak that clicking on a req on the tree from external credit and making it go away doesn't really make sense (unless that's a bug) since the course is still on the schedule. Other than that, everything looks really good! I think the bug is small enough that we can still just launch right now.

frontend/src/components/App.js
churichard: The name is retrieved from the database.

frontend/src/components/ExternalCreditForm.js
churichard: I think this is just a bug in GitHub's syntax highlighting. The code runs fine and there are no errors.

frontend/src/styles/Courses.css
churichard: I think it's an improvement over before as well, but I mainly just needed to change the color to something else in order to accommodate external credits. I'll keep thinking about ways to improve our color scheme (for example, now external credits stand out in the schedule more than I'd like them to).

churichard:
Clicking on the external credit in the requirements pane and having it go away is definitely a bug. Good catch, guys! I've fixed it and also addressed Barak's comments.

bnehoran (commented):
Looks good!

churichard (approved):

PR #243, opened 2019-04-11
Improve AP credits discoverability

danielleung27:
Improve AP credits discoverability


bnehoran (approved):
Yeah, I think this might help, but it might not be enough. It seems like if it were just the text, then people would be curious and try to click on it to see what happens. It could be that people just don't notice the button. 
So maybe we should replace the GUI element for the page switching with something that stands out more.

PR #245, opened 2019-08-13
Fix scraper to use webfeeds for distribution area

danielleung27:
Fix scraper to use webfeeds for distribution area

### DIFF
--- a/tigerpath/scraper/scrape_parse.py
+++ b/tigerpath/scraper/scrape_parse.py
@@ -9,8 +9,117 @@ Procedure:
 from lxml import etree
 from html.parser import HTMLParser
 from urllib.request import urlopen
-from bs4 import BeautifulSoup
 import re
+
+DEPTS = {
+    "AAS": "African American Studies",
+    "AFS": "African Studies",
+    "AMS": "American Studies",
+    "ANT": "Anthropology",
+    "AOS": "Atmospheric & Oceanic Sciences",
+    "APC": "Appl and Computational Math",
+    "ARA": "Arabic",
+    "ARC": "Architecture",
+    "ART": "Art and Archaeology",
+    "ASA": "Asian American Studies",
+    "AST": "Astrophysical Sciences",
+    "ATL": "Atelier",
+    "BCS": "Bosnian-Croatian-Serbian",
+    "CBE": "Chemical and Biological Engr",
+    "CEE": "Civil and Environmental Engr",
+    "CGS": "Cognitive Science",
+    "CHI": "Chinese",
+    "CHM": "Chemistry",
+    "CHV": "Center for Human Values",
+    "CLA": "Classics",
+    "CLG": "Classical Greek",
+    "COM": "Comparative Literature",
+    "COS": "Computer Science",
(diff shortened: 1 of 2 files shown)


bnehoran (dismissed):

tigerpath/scraper/scrape_parse.py
```diff
@@ -32,21 +141,27 @@ def scrape_parse_semester(term_code):
...
+            ParseError("key " + key + " does not exist")
+        else:
+            return h.unescape(found.text)
 
     def get_current_semester():
         """ get semester according to TERM_CODE
         """
         #global CURRENT_SEMESTER
+        # for now hardwire the namespaces--too annoying
+        PTON_NAMESPACE = u'http://as.oit.princeton.edu/xml/courseofferings-1_4'
```
danielleung27, line 164: Why don't we just use namespace 1_5, does it break this function?

tigerpath/scraper/scrape_parse.py
```diff
@@ -220,4 +319,4 @@ def remove_namespace(doc, namespace):
             if elem.tag.startswith(ns):
                 elem.tag = elem.tag[nsl:]
 
-    return scrape_all()
```
danielleung27, line 223: get rid of this

bnehoran (approved):

tigerpath/scraper/scrape_parse.py
```diff
@@ -32,21 +141,27 @@ def scrape_parse_semester(term_code):
...
 
     CURRENT_SEMESTER = ['']
 
     h = HTMLParser()
 
-    def get_text(key, object):
-        return h.unescape(raise_if_none(object.find(key), "key " + key + " does not exist").text)
+    def get_text(key, object, fail_ok=False):
+        found = object.find(key)
+        if fail_ok and (found is None or found.text is None):
```
danielleung27, line 152: I think it's more readable like this
`if fail_ok:`
____`if(found is None or found.text is None):`
________`return found`
____`else:`
________`return h.unescape(found.text)`
`else:`
 ____`return ParseError("key " + key + " does not exist")`

tigerpath/scraper/scrape_parse.py
bnehoran: Yeah, updating to 1.5 breaks it. Not sure why.

tigerpath/scraper/scrape_parse.py
bnehoran: That doesn't work in the case of `fail_ok=False`

tigerpath/scraper/scrape_parse.py
bnehoran: Shouldn't there be a newline at the end of python files? I thought that was the standard.
I know it's not related to the PR, but also it doesn't hurt.

danielleung27: oh ok, thats fine then. i thought it was an accident

tigerpath/scraper/scrape_parse.py
danielleung27: oh wait i see, my bad.

PR #261, opened 2019-08-27
Add course info and eval links to course popovers

bnehoran:


frontend/src/Popover.js
```diff
@@ -29,6 +33,26 @@ export function addPopover(course, courseKey, semIndex) {
     content += "<div class='popover-warning'>This course has previously only been offered in the Spring.</div>";
   }
 
+  let courseId = course["id"];
+  // TODO: Don't hardcode the semester list
+  let courseSemList = ['f19'];
```
bnehoran: This should not be hardcoded, especially since some courses don't appear that semester, but when I use
```suggestion
  let courseSemList = course["semester_list"];
```
I get `undefined` for some reason.

@danielleung27 maybe you can figure out how to access the semester list for the course.

frontend/src/Popover.js
```diff
@@ -29,6 +33,26 @@ export function addPopover(course, courseKey, semIndex) {
...
+  let termCode = convertSemToTermCode(courseSemList[courseSemList.length - 1]);
+  let courseInfoLink = BASE_COURSE_OFFERINGS_URL + "?courseid=" + courseId + "&term=" + termCode;
+  let courseEvalLink = BASE_COURSE_EVAL_URL + "?terminfo=" + termCode + "&courseinfo=" + courseId;
+
+  content += "<div className='search-card-links'>"
+  content += "<a href=" + courseInfoLink + " target='_blank' rel='noopener noreferrer'> "
+  // TODO: replace text link with actual icon button. Doesn't work for some reason
+  content += "🛈"  // i in circle symbol
+  // content += "<i className='fas fa-info-circle fa-lg fa-fw course-info'></i>"
```
bnehoran: I wasn't able to get the `fa-info-circle` symbol to show up here, so I temporarily placed a unicode "info" symbol instead.

frontend/src/Popover.js
```diff
@@ -29,6 +33,26 @@ export function addPopover(course, courseKey, semIndex) {
...
+  content += "</a> "
+  content += "<a href=" + courseEvalLink + " target='_blank' rel='noopener noreferrer'>"
+  // TODO: replace text link with actual icon button. Doesn't work for some reason
+  content += "📊"  // bar graph symbol
+  // content += "<i className='fas fa-chart-bar fa-lg fa-fw course-eval' />"
```
bnehoran: Same here with the unicode bar graph symbol.
@danielleung27 need your help

frontend/src/Popover.js
churichard: Try using `class` instead of `className`. I think this is rendered directly into the DOM, which uses `class`. In React JSX, though, `className` is used.

It's a little confusing because the way this popover is coded is bad style. It should technically be refactored into a React component.

frontend/src/Popover.js
churichard: I think `semester_list` is a property that exists only in the search results course card (`SearchCard`) and not the schedule course card (`CourseCard`). If you want to access it, you can modify the endpoint `get_schedule` in the backend to return it in each course (which is called here to get the schedule course cards: https://github.com/TigerPathApp/tigerpath/blob/38c709d166ff40b7c4030f31a18ec14999aa0a19/frontend/src/components/Schedule.js#L21-L29).

PR #293, opened 2020-06-24
Update Computer Science - AB major (COS-AB)

bcatoto:


tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -237,16 +237,25 @@
...
+      "course_list": [
+        "COS 397",
+        "COS 398"
+      ]
```
bnehoran: Are you sure about this? Was there a policy change that you know of?
As far as I know, only BSE students enroll in COS 397 and 398, right?

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -76,15 +77,13 @@
...
...Physics or ORF course that does not duplicate COS content, Philosophy 312, MAE 345, CHM 303(CHM 303 is now renumbered to CHM 301 as of 8/27/2018), ECO 312, MOL 437/NEU 437, NEU 330, MUS 315, and ECO 326 – Economics of the Internet count as departmentals. On occasion, special courses with computational content from other departments may fall in to this category. Note that it is very, very, very rare for courses outside the computer science department that are not cross-listed with the computer science department to count as track-specific (ie: Theory/Systems/Applications) departmentals. Hence, in general, you will not be able to count more than 2 courses outside of the computer science department as COS departmentals.\n\nSome regular courses are COS departmentals but do not count in any specific track, such as COS 351 \"Information technology and public policy\". COS 495, Special Topics, is a departmental but may be associated with different tracks (or no track) depending on content. Other one-time courses may also be departmentals and may count in one of the tracks depending on level and content. Check with your advisor or the undergraduate coordinator for up-to-date information.",
```
bnehoran, line 81: You're just removing the extra spaces here, right? There are no other differences?

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -76,15 +77,13 @@
...
       "req_list": [
         {
           "name": "Theory",
           "max_counted": 1,
           "min_needed": 2,
-          "explanation": "Courses in this track explore the theoretical underpinnings of computing systems.  They all use mathematical proofs for rigorous development of the area of study.\n\nTheory courses:\n• 340 Reasoning about Computation\n• 423 Theory of Algorithms\n• 433**Cryptography\n• 445 Networks, Economics and Computing\n• 451 Computational Geometry\n• 487 Theory of Computation\n• 488 Introduction to Analytic Combinatorics\n• 510* (441) Programming Languages\n• 511 Theoretical Machine Learning\n• 516* Reasoning About Software\n• 533**  Advanced Cryptography\n\n*If you take COS 510 (or 441) and COS 516 only one will count as a theory requirement.\n**If you take COS 433 and COS 533 only one will count as a theory requirement.",
+          "explanation": "Courses in this track explore the theoretical underpinnings of computing systems. They all use mathematical proofs for rigorous development of the area of study.",
```
bnehoran: At least leave in the note about courses that cannot both count. Otherwise, users would get confused as to why TigerPath is not counting them.

bnehoran: Also, there is no harm in having the course list itself in the explanation, though yes, it is a bit redundant, so if you think it's extra, I have no qualms about removing it.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -128,7 +129,7 @@
           "name": "Systems",
           "max_counted": 1,
           "min_needed": 2,
-          "explanation": "Courses in this track study the design and implementation of the foundational hardware and software that constitute a modern computing environment.\n\nSystems courses:\n• 306 (ELE 206) Introduction to Logic Design\n• 316 Principles of Computer System Design\n• 318 Operating Systems\n• 320 Compiling Techniques\n• 333 Advanced Programming Techniques\n• 375 (ELE 375) Computer Architecture and Organization\n• 418 Distributed Systems**\n• 425 Database and Information Management Systems\n• 461 Computer Networks*\n• 463 Wireless Networks\n• 475 (ELE 475) Computer Architecture\n• 518 Advanced Computer Systems**\n• 561 Advanced Computer Networks*\n\n*If you take COS 461 and  COS 561 only one will count as a systems requirement.\n**If you take COS 418 and  COS 518 only one will count as a systems requirement.",
+          "explanation": "Courses in this track study the design and implementation of the foundational hardware and software that constitute a modern computing environment.",
```
bnehoran: Same here. At least leave in the note about courses that cannot both count.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -173,7 +174,7 @@
...
           "max_counted": 1,
           "min_needed": 2,
-          "explanation": "Courses in this track treat a broad range of topics.\n\nApplications courses:\n• 314 (MUS 314) Computer and Electronic Music through Programming, Performance, and Composition (with programming precept)\n• 323 (ORF 363)  Computing and Optimization \n• 324 Introduction to Machine Learning\n• 326 Functional Programming\n• 343 Algorithms for Computational Biology\n• 360 (PSY 360) Computational Models of Cognition\n• 401 (TRA 301) Introduction To Machine Translation\n• 402 Machine Learning and Artificial Intelligence\n• 424 Fundamentals of Machine Learning\n• 426* Computer Graphics\n• 429 Computer Vision\n• 432 Information Security\n• 435 Information Retrieval, Discovery and Delivery\n• 436 Human-Computer Interface Technology\n• 455 (MOL 455) Introduction to Genomics and Computational Molecular Biology\n• 484 Natural Language Processing\n• 485 Neural Networks\n• 526* Advanced Computer Graphics\n\n*If a student takes both COS 426 and 526, only one will count as an applications requirement.",
+          "explanation": "Courses in this track treat a broad range of topics.",
```
bnehoran: Same here.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -4,26 +4,31 @@
...
+  "explanation": "Computer science concentrators must take 8 departmental courses (departmentals) and do independent work. The minimum amount of independent work depends on whether you are an AB candidate (4 semesters minimum) or a BSE candidate (1 semester minimum). Departmental courses are any computer science course at the 300 or 400 level (except 397, 398, 497 and 498) and a specific set of courses in other departments (enumerated below). There is also a distribution requirement within the 8 departmental courses: you must take at least 2 Systems departmentals, at least 2 Theory departmentals, and at least 2 Applications departmentals. (Notice that leaves you 6 departmentals that are constrained and 2 departmentals that are unconstrained.)\n\nBSE candidates do independent work by signing up for COS 397 or 398 in the junior year, or COS 497 or 498 in the senior year. You encouraged to do more than one semester of independent work; the second semester of independent work may be counted as a departmental.\n\nAB candidates do 2 one-semester junior independent work projects and a full year senior thesis. Numbered courses are not used to designate AB independent work.",
```
bnehoran: It actually _is_ called `"description"` here, according to the [schema](https://github.com/TigerPathApp/tigerpath/blob/master/tigerpath/majors_and_certificates/scripts/schema.json). 
It's a description of the major.

In case you're curious, there were initially three different fields that ended up sort of converging on each other:
* `explanation` for a requirement, initially intended as a sort of guide for someone looking through the requirements file, almost like a comment to help someone reviewing the code.
* `description` _for a subrequirement_, intended as a concise few-word, one-line description of the requirement to be displayed to the user (now deprecated)
* `description` _of the major_, a sort of elevator pitch from the department website for why users should choose this major

We at some point realized that we could display a lot more details in the popovers for each subrequirement, so eventually we ended up displaying the `explanation` field as well as the `description`, but then the `description` field became redundant, so we deprecated it, moving everything into the explanation.
So the first two merged.

Meanwhile the third one in the list, the `description` for the overall major, still exists, though if you think it's confusing, then maybe we should change it all to be called `explanation`.
I guess I'd agree that it's now gotten to where it's practically identical to the `explanation` for subrequirements, which means we might as well merge them.

But it should be a change that's standardized across all majors, and we would need to update the schema and documentation as well, so that should happen in its own separate pull request.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -32,15 +37,12 @@
       "max_counted": 1,
       "min_needed": "ALL",
       "explanation": "The prerequisites for AB candidates are MAT 103, 104, and 202 or 204 or 217; COS 126, COS 217, and 226.\n\nStudents should plan to take both 217 and 226 before the junior year. One or both of these are required prerequisites for all later computer science courses.",
-      "double_counting_allowed": false,
-      "pdfs_allowed": false,
```
bnehoran, line 36: Why is this being removed?
Was there a departmental policy change behind this allowing PDFing the prereqs? (That is, of course, besides that for the COVID19 pandemic, which says that all Spring 2020 courses can be PDFed, but that's a separate matter.)

We don't currently display this information, by the way, but we might in a future update.
Of course, we would never actually check whether or not a student in fact chooses to PDF a course. (How could we? We don't and can't see their grades; nor do we want to.) We would most likely just display that information to them in the popover. But then why not leave it in, so that it's there if/when we need it.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -76,15 +77,13 @@
...
```
bnehoran, line 81: Same here. See my comment on line 36/39 above. Why is this removed?

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
```diff
@@ -237,16 +237,25 @@
             "NEU 330",
             "MUS 315",
             "ECO 326"
+          ],
+          "excluded_course_list": [
+            "COS 397",
+            "COS 398",
+            "COS 497",
+            "COS 498"
```
bnehoran, line 247: Nice!

bnehoran (changes requested):

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
bcatoto: I'd thought the default was the double-counting was false and PDF allowed is also false so I removed them here (and in other majors) because I thought it was redundant and slightly confusing when I was trying to keep track of the double-counting. If you'd prefer to have it back, let me know.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
bcatoto: Yeah, just a small detail I noticed.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
bcatoto: COS 397 and COS 398 are actually optional to fulfill Independent Work requirements for COS-AB but I didnt' know about the optional track trick because when I made Independent Work null, the root wouldn't turn green.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
bnehoran: Yeah, you're right, `double_counting_allowed` is false by default, so no need for it to be listed explicitly unless it's true, so I support removing that.

`pdfs_allowed`, on the other hand, doesn't really have a default yet as far as I know, because we haven't really settled yet on how to interpret and display it. Right now, if it's not listed, it mostly just means unspecified, I think.
It's not required to be listed, but if it's already there, no need to remove it, in my opinion.
But, yeah, I'm not hung up on it, honestly. It's not an issue either way, since it has no real effect at the moment, so I'll resolve this comment.

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
bnehoran: Cool, no problem. Just wanted to be sure I'm not missing a content change hidden somewhere in there!

tigerpath/majors_and_certificates/majors/COS-AB_2018.json
bnehoran: Could you link to a departmental source that says that this is the case? Because as far as I can tell, the department webpage, and the course listings themselves, both very clearly say "Open to BSE COS Concentrators Only".

bnehoran (dismissed):

bnehoran (approved):

PR #294, opened 2020-06-24
Update Computer Science - BSE major (COS-BSE)

bcatoto:


tigerpath/majors_and_certificates/majors/COS-BSE_2018.json
```diff
@@ -32,35 +37,24 @@
...
-          "course_list": [
-            "COS 126",
-            "COS 217",
-            "COS 226"
-          ]
-        }
+      "course_list": [
+        "COS 126",
+        "COS 217",
+        "COS 226"
```
bnehoran, line 45: The reason for having it structured this way is so that the AB and BSE COS requirements have parallel structure.
That makes it easier to, say apply the same patch to both of them, or diff between them.
Do you think it's confusing for users such that it's worth relaxing that restraint?

tigerpath/majors_and_certificates/majors/COS-BSE_2018.json
```diff
@@ -191,13 +186,13 @@
...
-              "explanation": "Any computer science (COS) course at the 300 or 400 level (except independent work) counts as a computer science departmental.\nIn addition, any 300- or 400- level Math or ELE or Physics or ORF course that does not duplicate COS content, Philosophy 312, MAE 345, CHM 303(CHM 303 is now renumbered to CHM 301 as of 8/27/2018), ECO 312, MOL 437/NEU 437, NEU 330, MUS 315, and ECO 326 – Economics of the Internet count as departmentals.  On occasion, special courses with computational content from other departments may fall in to this category.",
+              "explanation": "Any computer science (COS) course at the 300 or 400 level (except independent work) counts as a computer science departmental.\nIn addition, any 300- or 400- level Math or ELE or Physics or ORF course that does not duplicate COS content, Philosophy 312, MAE 345, CHM 303(CHM 303 is now renumbered to CHM 301 as of 8/27/2018), ECO 312, MOL 437/NEU 437, NEU 330, MUS 315, and ECO 326 – Economics of the Internet count as departmentals. On occasion, special courses with computational content from other departments may fall in to this category.",
               "course_list": [
```
bnehoran, line 198: Should add Independent work courses COS 397/398/497/498 as excluded course from this list so that they are forced to fall in the requirement below it.
We just didn't have `excluded_course_list` when we first encoded this.

bnehoran (changes requested):
Almost all the same comments from https://github.com/TigerPathApp/tigerpath/pull/293 also apply here, to the same locations. I've only added two comments on top of those here, which are unique to the BSE version.

A neat trick is that if you keep them parallel in their structure, you can often take the patch that you apply to one and just apply to the other in almost the same stroke.

bnehoran (approved):

PR #296, opened 2020-06-24
Update Economics major (ECO)

bcatoto:


tigerpath/majors_and_certificates/majors/ECO_2018.json
```diff
@@ -25,227 +25,261 @@
...
+          "explanation": "To enter the department, a student must complete, by the end of sophomore year, the prerequisite courses ECO 100, ECO 101, ECO 202 (or equivalent), and MAT (175 or equivalent), earning a letter grade of C or better in each course. ORF 245 can be substituted for ECO 202. POL 345 followed by POL 346 can also be used to satisfy our statistics pre-requirement. (POL 345 alone is not sufficient).",
           "req_list": [
             {
               "name": null,
-              "max_counted": 1,
-              "min_needed": 1,
-              "description": null,
-              "explanation": "",
+              "max_counted": null,
+              "min_needed": "ALL",
```
bnehoran: This just makes the counting more responsive. The logic is the same since the same thing is enforced by the `"min_needed": 3` on line 34.
```suggestion
              "max_counted": 2,
              "min_needed": null,
```

tigerpath/majors_and_certificates/majors/ECO_2018.json
```diff
@@ -25,227 +25,261 @@
...
+            },
+            {
+              "name": null,
+              "max_counted": 1,
+              "min_needed": "ALL",
+              "explanation": null,
+              "course_list": [
+                "MAT 203",
+                "MAT 204"
+              ]
```
bnehoran: Aren't students allowed to take a combination like MAT 203 with MAT 202?
That should be fine, right?
```suggestion
            {
              "name": null,
              "max_counted": 1,
              "min_needed": "ALL",
              "explanation": null,
              "req_list": [
                {
                  "name": null,
                  "max_counted": 1,
                  "min_needed": null,
                  "explanation": null,
                  "course_list": [
                    "MAT 201",
                    "MAT 203",
                  ]
                },
                {
                  "name": null,
                  "max_counted": 1,
                  "min_needed": null,
                  "explanation": null,
                  "course_list": [
                    "MAT 202",
                    "MAT 204"
                  ]
                }
              ]
```

tigerpath/majors_and_certificates/majors/ECO_2018.json
```diff
@@ -25,227 +25,261 @@
...
+            "WWS 582C",
+            "WWS 582F"
+          ]
+        }
+      ]
+    },
+    {
+      "name": "Tracks",
       "max_counted": 1,
       "min_needed": 1,
```
bnehoran: This structure, with the redundant "Economics" track, technically works, though I agree that it's a bit odd. The structure you had before was nicer, but of course, it was causing issues.

The issue here is that the verifier currently doesn't bother checking requirements that don't contribute to anything (the thought when designing it was: well, why should it?), which is I guess why it gets left unchecked if you just do the simpler thing of promoting the PPE requirement and setting "max_counted" to 0.
I guess I didn't really anticipate having a completely useless, uproductive requirement (in the sense of not helping to satisfy the overall major) that should nevertheless still be checked, but thinking about it now, that doesn't seem as weird as I had thought, and so maybe it should be reevaluated.

One funny hack you could use is to set "max_counted" to some really small non-zero value like 0.000000000001, which is non-zero (forcing the verifier to check the requirement), but at the same time not large enough to make any difference whether it's satisfied or not.
It's definitely a weird hack, but it does the job of letting the verifier check the requirement but also not passing anything up.

So I would use that here and then drop the redundant Economics track. Perhaps also rename this to "Optional Tracks" to make it clear that it's optional.

It's important to properly comment the hack though, so that it's clear why it has such a weird value, and also so we know to remove it once we upgrade the verifier.

```suggestion
      "name": "Optional Tracks",
      # the following line is a hack to allow for an unproductive requirement that still gets checked. See discussion at PR #296 
      "max_counted": 0.000000000001,
      "min_needed": 0,
```

tigerpath/majors_and_certificates/majors/ECO_2018.json
```diff
@@ -25,227 +25,261 @@
...
-      "explanation": "The Department offers a Program in Political Economy for Economics concentrators interested in studying issues at the intersection of these two fields. Program participants meet all the general requirements of Economics concentrators, and, in addition: (1) complete two Politics courses at any level before entering the program; (2) complete three 300-level (or higher) Politics courses during junior and senior years, and (3) conduct junior independent work with a political economy component. Politics courses approved as Economics cognates may also count toward the program’s requirements. Admission to the program takes place at the beginning of the junior year, in consultation with the Political Economy Adviser in Economics. A student who participates in this program is an Economics concentrator and is subject to all the requirements of the Economics department.  Concentrators who successfully complete the program’s requirements will receive a departmental attestation.",
+      "explanation": null,
       "req_list": [
         {
-          "name": "Economics (Standard)",
+          "name": "Economics",
```
bnehoran: If adopting the hack (see comment on line 213), you can then remove this redundant track.

bnehoran (changes requested):

bnehoran (approved):

PR #297, opened 2020-06-25
Update Ecology and Evolutionary Biology major (EEB)

bcatoto:


tigerpath/majors_and_certificates/majors/EEB_2018.json
```diff
@@ -55,37 +56,63 @@
...
+            },
+            {
+              "name": null,
+              "max_counted": 2,
+              "min_needed": null,
+              "explanation": null,
+              "course_list": [
+                "CHM 215"
+              ]
+            }
```
bnehoran: Even though `max_counted` is 2, you still need to add the course twice to get there.

A cool trick is to duplicate the hidden subrequirement and then set its parent's `double_counting_allowed` to true. (In fact why not set it for the full Prerequisites requirement, as they don't overlap with departmentals anyway?)
Doing so lets the course count as double. The only downside is that it also appears doubled in the requirements tree, but that's probably not an issue unless you think it would confuse people.
```suggestion
            {
              "name": null,
              "max_counted": 1,
              "min_needed": null,
              "explanation": null,
              "course_list": [
                "CHM 215"
              ]
            },
            {
              "name": null,
              "max_counted": 1,
              "min_needed": null,
              "explanation": null,
              "course_list": [
                "CHM 215"
              ]
            },
```

Up to you if you think this is inelegant, maybe there's another way.

tigerpath/majors_and_certificates/majors/EEB_2018.json
```diff
@@ -26,18 +26,19 @@
       "max_counted": 1,
       "min_needed": "ALL",
       "explanation": "The Department of Ecology and Evolutionary Biology requires the following prerequisites, which should normally be completed by the end of sophomore year.",
+      "completed_by_semester": 4,
```
bnehoran, line 29: See my comment on line 91.
```suggestion
      "completed_by_semester": 4,
      "double_counting_allowed": true,
```

tigerpath/majors_and_certificates/majors/EEB_2018.json
```diff
@@ -97,113 +124,117 @@
...
           "course_list": [
             "SML 201",
-            "ECO 202",
             "ORF 245",
-            "POL 345",
             "PSY 251",
+            "POL 345",
+            "SOC 301",
+            "ECO 202",
+            "WWS 220",
```
bnehoran: This course doesn't exist. I think it's a typo and they meant WWS 200, which is "Statistics for Social Science"
```suggestion
            "WWS 200",
```
I mean, we could check with the department if you really want to be sure.

tigerpath/majors_and_certificates/majors/EEB_2018.json
```diff
@@ -97,113 +124,117 @@
...
             "WWS 332"
```
bnehoran, line 143: And then this course hasn't been offered in ages, but haha I guess it's fine to leave it in in case it reappears somehow.

bnehoran (changes requested):
Looks really good!

Just a few comments.

bnehoran (approved):

PR #307, opened 2020-06-26
Update Neuroscience major (NEU)

bcatoto:


bnehoran (changes requested):
I agree on principle that 2 spaces looks better.
However, if you want to update the tab spacing, could you maybe split it into two different commits: one which does only the spacing changes, and another which makes actual material changes?
Otherwise it's impossible to review, because the diff just becomes unreadable.

bcatoto:
I created a commit with just the content and no spacing change and then another with the spacing updated.

bnehoran:
Not really sure what's going on here, but it still looks like every one of the commits has a diff that covers the whole file because of the spacing changes.
What would be nice is having one commit that I can look at (or even a sequence of commits) where the only lines in the diff are for content changes.
Otherwise, it's just impossible to decipher what's going on, since it's often not even matching the correct pairs of lines in the diff.

PR #310, opened 2020-06-26
Update Physics major (PHY)

bcatoto:


tigerpath/majors_and_certificates/majors/PHY_2018.json
```diff
@@ -214,7 +214,7 @@
           ]
```
bnehoran, line 214: This is, of course, not related to your changes, but do you think we should leave a note to the users on the "Additional Physics" requirement that not all of the courses in the list are valid to satisfy the requirements?
The app does not check which department is listed first, so crosslisted courses will appear too.

tigerpath/majors_and_certificates/majors/PHY_2018.json
```diff
@@ -116,13 +116,13 @@
           "name": "Math",
           "max_counted": 1,
           "min_needed": "ALL",
-          "explanation": "Mastery of calculus to the level of MAT 103 or above, or advanced placement (an AP score of 5 on the AB test or an AP score of 4 on the BC test);",
+          "explanation": "Prerequisites for concentration in physics are the following five courses: PHY 103-104, PHY 205, and MAT 203-204. MAT 203 may be replaced by MAT 201 or MAT 218. MAT 204 may be replaced by MAT 202 or MAT 217.",
```
bnehoran, line 119: Wow good catch! I don't know why I didn't notice that before.

tigerpath/majors_and_certificates/majors/PHY_2018.json
```diff
@@ -263,18 +263,16 @@
...
-      "description": "Independent Work",
+      "explanation": "<i>Editor's Note: Concentrators of this major do not enroll in any numbered course for this requirement, so TigerPath does not check it.</i>\n\nIn addition to the coursework carried out during the junior year, the student is required to complete two junior papers, each of which is on a research topic of current interest. The purpose of the papers is to give students exposure to how physics research is actually performed by immersing them in journal, as opposed to textbook, literature. Each paper is written in close consultation with a faculty adviser, who is typically performing research in the subject area of the paper. A junior paper may serve as a preliminary investigation of a senior thesis topic. Junior independent work may also be satisfied with a short experimental project.",
```
bnehoran, line 268: Thanks for finding these mistakes! It seems like this was copied from the EEB major without modification. Good that you noticed!

bnehoran (dismissed):
Thanks for finding and fixing these mistakes!
They must have slipped through the cracks last time, so I'm really glad that you noticed them. Thank you!

Just one minor comment about maybe adding an editor's note to one of the requirements. It's not related to any of your changes, but I just happened to notice it as I was looking over it.
I you get the chance and want to add it in, then great. It's not a hold-up though.

Also, just a note for the future: it seems like you have many inconsequential changes in this PR (changing `""` to `null` for example). I think I'd mentioned this to you, that in such a case, it often makes it a bit easier to review if you separate those into a separate properly labeled commit. It's not a big deal, of course, but just wanted to bring it to your attention.

Good work!

tigerpath/majors_and_certificates/majors/PHY_2018.json
bcatoto: Yeah, I think that would be useful.

bnehoran (approved):
Looks good!

PR #312, opened 2020-06-26
Update Psychology major (PSY)

bcatoto:


tigerpath/majors_and_certificates/majors/PSY_2018.json
```diff
@@ -111,12 +116,28 @@
           "name": "Other Departmentals",
           "max_counted": 4,
           "min_needed": 4,
-          "explanation": "",
-          "course_list": [
-            "PSY 2*",
-            "PSY 3*",
-            "PSY 4*",
-            "PSY 5*"
+          "explanation": "Of the remaining departmental courses, all must be 200-level or higher and at least four must be 300-level or higher. Students can take up to two cognate courses from other departments, approved by the Psychology Departmental Representative, to count as departmental courses.",
```
bnehoran, line 122: Something's not quite right here. If at least 4 of 4 courses muse be 300-level or higher, that's all of them. What am I missing?

tigerpath/majors_and_certificates/majors/PSY_2018.json
```diff
@@ -75,20 +80,20 @@
...
             {
               "name": null,
-              "max_counted": 3,
+              "max_counted": null,
               "min_needed": null,
-              "explanation": "",
+              "explanation": null,
               "course_list": [
                 "PSY 252",
                 "PSY 255",
```
bnehoran, line 102: I understand that this was partly already weird before your changes, but what happens now if those three courses were already taken as prerequisites? Then this requirement wouldn't be satisfiable as it's coded here, right?

bcatoto: I'm not quite sure I understand the concern. The way it's coded right now is that these courses also double count as fulfilling a prerequisite and the required departmentals.

image

tigerpath/majors_and_certificates/majors/PSY_2018.json
bcatoto: You're right I think I thought of PSY 300 as part of the 300-level courses but it shouldn't be.

tigerpath/majors_and_certificates/majors/PSY_2018.json
bnehoran: Yeah, you're right. I confused myself there.

bnehoran (approved):

PR #313, opened 2020-06-26
added Religion major (REL)

bcatoto:


tigerpath/majors_and_certificates/majors/REL_2018.json
```diff
@@ -0,0 +1,221 @@
...
+        },
+        {
+          "name": "Department Courses",
+          "max_counted": 2,
+          "min_needed": null,
+          "explanation": null,
+          "course_list": [
+            "REL *"
+          ],
+          "excluded_course_list": [
```
bnehoran, line 175: There is currently a problem when searching for requirements that have excluded courses, since they still show up in the search. Essentially the `excluded_course_list` is used when checking which courses satisfy the requirement, but at the moment, it's not passed as a parameter to the search functionality, so I'll look into fixing that.
This is just to say, it's causing problems with this requirement, but the issue isn't here.

tigerpath/majors_and_certificates/majors/REL_2018.json
```diff
@@ -0,0 +1,221 @@
...
+            "REL 250/AAS 250",
+            "REL 256/AAS 256",
+            "REL 259/GSS 278/LAS 259",
+            "REL 271/AMS 341",
+            "REL 275/LAS 275",
+            "REL 357/HIS 310",
+            "REL 359/LAS 388",
+            "REL 360/GSS 360/AMS 369",
+            "REL 371",
+            "REL 373",
```
bnehoran, line 95: The department says that this course might count for a different area requirement depending on the semester and the subject covered in that semester.
Because of this, I think it should probably appear either under all the areas (letting the user decide where to settle it) or none of them (essentially indicating they shouldn't assume that it will count anywhere).
Either way, we could also add the note from the [department](https://religion.princeton.edu/undergraduate-program/major-requirements/) about this course to the end of the "Subfields" explanation.

bnehoran (approved):
Looks great! Just one small thing.

tigerpath/majors_and_certificates/majors/REL_2018.json
bcatoto: I got the subfield course lists from the department's past semester course lists, which includes the subfields the courses fall under for that semester. So far REL 373 has only been categorized as part of the Religion in America subfield. I think to keep TigerPath accurate, it can be added to the other subfields when there is a semester where it's categorized in another subfield.

bnehoran: Fair enough. We just need to keep an eye out for it then, I guess.

PR #315, opened 2020-06-26
Update Sociology major (SOC)

bcatoto:


tigerpath/majors_and_certificates/majors/SOC_2018.json
```diff
@@ -36,37 +36,62 @@
...
+                "SOC 302",
+                "POL 345/SOC 305"
+              ]
+            }
           ]
         }
       ]
     },
     {
       "name": "Independent Work",
```
bnehoran: A couple things about this:
- Isn't `POL 345/SOC 305` now being replaced by `SOC 301` ?
- This would then be redundant with the "Required" req, so we could just revert it back to the usual `no_req`, right?

This was probably a mistake when originally encoding this major, but I don't think the department is saying that these two courses constitute a new separate requirement, but rather that students use what they learn from those two courses (which are in the list of required courses above) as tools for their independent work.
So no need to require them again here explicitly.

tigerpath/majors_and_certificates/majors/SOC_2018.json
```diff
@@ -36,37 +36,62 @@
...
+                "SOC 4*",
+                "SOC 5*"
+              ],
+              "excluded_course_list": [
+                "SOC 300",
+                "SOC 301",
+                "SOC 302",
+                "POL 345/SOC 305"
+              ]
+            }
```
bnehoran, line 122: Really nice!

tigerpath/majors_and_certificates/majors/SOC_2018.json
```diff
@@ -36,37 +36,62 @@
...
           "max_counted": 1,
           "min_needed": "ALL",
-          "explanation": "",
+          "explanation": "Within the nine sociology courses, the following four must be taken: SOC 101: Introduction to Sociology (taught in the fall and the spring and can be taken anytime, even after taking mid- and upper-level sociology courses), SOC 300: Claims and Evidence (taken in the Fall of the junior year to prepare students for the Junior Paper), SOC 301: Statistical Methods in Sociology, SOC 302: Sociological Theory.",
+          "completed_by_semester": 6,
           "course_list": [
             "SOC 101",
             "SOC 300",
-            "POL 345/SOC 305",
+            "SOC 301",
```
bnehoran: Good update.
I think the 3rd point [here](https://sociology.princeton.edu/undergraduate-program/prospective-majors/faqs) might still apply, though, which seems to imply that now, instead of SOC 301, they can still take POL 345/SOC 305, or ORF 245/EGR 245, PSY 251, or WWS 200.

So maybe it would be best to split this course out into a separate req which allows any one of those courses to satisfy it, with that FAQ answer as the explanation.

tigerpath/majors_and_certificates/majors/SOC_2018.json
```diff
@@ -6,7 +6,7 @@
   "year": 2018,
   "urls": [
     "https://ua.princeton.edu/academic-units/department-sociology",
-    "https://sociology.princeton.edu/"
+    "https://sociology.princeton.edu/undergraduate-program/current-majors/requirements"
```
bnehoran: Undergraduate handbook
```suggestion
    "https://sociology.princeton.edu/undergraduate-program/current-majors/requirements",
    "https://sociology.princeton.edu/sites/sociology/files/undergraduate_handbook_201920.pdf"
```

bnehoran (changes requested):

tigerpath/majors_and_certificates/majors/SOC_2018.json
bnehoran: Ah indeed, [this page](https://sociology.princeton.edu/undergraduate-program/sophomores) confirms it, and is probably a better source, since it seems to be more up to date.

bnehoran (approved):

PR #318, opened 2020-06-26
Add Art and Archaeology major (ART)

bcatoto:


tigerpath/majors_and_certificates/majors/ART_2018.json
```diff
@@ -0,0 +1,272 @@
+{
+  "type": "Major",
+  "name": "Art and Archaeology",
+  "code": "ART",
+  "degree": "AB",
+  "year": 2018,
+  "urls": [
+    "https://ua.princeton.edu/academic-units/department-art-and-archaeology",
+    "https://artandarchaeology.princeton.edu/"
```
bnehoran:
```suggestion
    "https://artandarchaeology.princeton.edu/",
    "https://cms.artandarchaeology.princeton.edu/media/files/2019a-and-augradhandbook-1.pdf"
```

tigerpath/majors_and_certificates/majors/ART_2018.json
```diff
@@ -0,0 +1,272 @@
...
+                  "course_list": [
+                    "ART 4*",
+                    "ART 5*"
+                  ]
+                },
+                {
+                  "name": "Additional Courses",
+                  "max_counted": 1,
+                  "min_needed": 6,
+                  "explanation": "Students must also take at least one course in each of the following three areas: Group 1 (ancient), Group 2 (medieval/early modern), and Group 3 (modern/contemporary). Thematic courses spanning more than one area will be allocated to a distribution area on a case-by-case basis. ",
```
bnehoran: Maybe we can contact the department to get more specific lists for the three groups. Fine for now.

bnehoran: Remove trailing space
```suggestion
                  "explanation": "Students must also take at least one course in each of the following three areas: Group 1 (ancient), Group 2 (medieval/early modern), and Group 3 (modern/contemporary). Thematic courses spanning more than one area will be allocated to a distribution area on a case-by-case basis.",
```

tigerpath/majors_and_certificates/majors/ART_2018.json
```diff
@@ -0,0 +1,272 @@
...
+          ]
+        },
+        {
+          "name": "Practice of Art",
+          "max_counted": 1,
+          "min_needed": "ALL",
+          "explanation": "Concentrators in the Practice of Art Track explore the traditions, thought processes, and methods of making visual art in connection with a general program of humanistic education. Courses are offered in painting, drawing, printmaking, sculpture, photography, film and video production, and film history and theory.",
+          "req_list": [
+            {
+              "name": "Prerequisites",
```
bnehoran, line 102: How come this doesn't have double counting enabled? 
Couldn't courses counting for prerequisites also count as departmentals?

Is it to prevent the same course counting for both the Visual Arts and Art and Archaeology subrequirements (if the course is cross-listed)?
In that case one possible fix would be to add an excluded course list to one of the two subrequirements in order to exclude courses that satisfy the other.
That would effectively choose for the user which category to put such course, though. Not sure if that could be problematic.

The longer term solution would might be to design a more sophisticated double-counting mechanism to handle these kinds of cases.

tigerpath/majors_and_certificates/majors/ART_2018.json
```diff
@@ -0,0 +1,272 @@
...
+                      "course_list": [
+                        "VIS 392"
+                      ]
+                    },
+                    {
+                      "name": "Art and Archaeology",
+                      "max_counted": null,
+                      "min_needed": null,
+                      "explanation": "<i>Editor's Note: This requirement is too broad, so TigerPath does not check it.</i>\n\nThe Department of Art and Archaeology courses must include: at least one course from Group 3 (19th century to the present); and two others, including at least one course from Group 1 (ancient) or Group 2 (medieval/early modern). The third course may be from any group.",
+                      "no_req": null
```
bnehoran: At least check for 3 "ART *" courses, I think, and let the user verify themselves that the courses should count in the specific groups. 
We can leave the editor's note. That's good.
Essentially, even if we can't verify it completely, if we can verify part of the requirement, in this case the number of courses, then why not do that.
It would also allow the user to add courses to the requirement which is a bit more satisfying than leaving it empty.

Also, as it is, we're checking for 7 courses, so adding this should allow us to check for all 10.

bnehoran: And, as with the comment on line 75, the best would be if we could get a list of courses in each group from the department.

tigerpath/majors_and_certificates/majors/ART_2018.json
```diff
@@ -0,0 +1,272 @@
...
+              "explanation": "A total of 10 courses, of which at least seven must be from the Program in Visual Arts and three must be from the Department of Art and Archaeology and taught by Art and Archaeology faculty.",
+              "req_list": [
+                {
+                  "name": "Visual Arts",
+                  "max_counted": 1,
+                  "min_needed": "ALL",
+                  "explanation": "The visual arts courses must include: studio courses in at least two different media; two studio courses at the 300 or 400 level; VIS 392 Issues in Contemporary Art; and VIS 416 Exhibition Issues and Methods or VIS 417 Fall Film Seminar. The Department of Art and Archaeology courses must include: at least one course from Group 3 (19th century to the present); and two others, including at least one course from Group 1 (ancient) or Group 2 (medieval/early modern). The third course may be from any group.",
+                  "req_list": [
+                    {
+                      "name": "VIS 416/VIS 417",
```
bnehoran: There seems to be an issue with forward slashes not being escaped properly, and this is causing the search for this requirement to fail. 
Eventually we should find and fix the underlying issue, of course, but for now, you can avoid the issue by slightly renaming the requirement.
Something like
```suggestion
                      "name": "VIS 416 or VIS 417",
```

tigerpath/majors_and_certificates/majors/ART_2018.json
```diff
@@ -0,0 +1,272 @@
...
+                  "explanation": null,
+                  "course_list": [
+                    "ART 100"
+                  ]
+                },
+                {
+                  "name": "Seminars",
+                  "max_counted": 1,
+                  "min_needed": 2,
+                  "explanation": "<i>Editor's Note: Many of these courses fall under this requirement but as TigerPath does not verify this.</i>\n\nTwo seminars at the 400- or 500-level",
```
bnehoran: It seems like this same editor's not has been copied to multiple different locations, all with an extra "as" included!

tigerpath/majors_and_certificates/majors/ART_2018.json
```diff
@@ -0,0 +1,272 @@
...
+                      "course_list": [
+                        "VIS 416",
+                        "VIS 417"
+                      ]
+                    },
+                    {
+                      "name": "Studio Courses",
+                      "max_counted": 2,
+                      "min_needed": 2,
+                      "explanation": "<i>Editor's Note: Most of these courses fall under this requirement but as TigerPath does not verify this.</i>\n\nTwo studio courses at the 300 or 400 level.",
```
bnehoran: as

tigerpath/majors_and_certificates/majors/ART_2018.json
```diff
@@ -0,0 +1,272 @@
...
+                  "req_list": [
+                    {
+                      "name": "Studio Courses",
+                      "max_counted": 1,
+                      "min_needed": 5,
+                      "explanation": "<i>Editor's Note: Most of these courses fall under this requirement but as TigerPath does not verify this.</i>\n\nThree studio courses at any level in at least two different media (screenwriting courses are accepted as a different medium from film production courses) and an additional two studio courses at the 300 or 400 level.",
```
bnehoran: as

tigerpath/majors_and_certificates/majors/ART_2018.json
bnehoran: Also, is it actually _most_?
Or is it rather "not all of these courses"?

tigerpath/majors_and_certificates/majors/ART_2018.json
bnehoran: Same here. Is it actually _most_?

bnehoran (changes requested):

tigerpath/majors_and_certificates/majors/ART_2018.json
bcatoto: I emailed the Art and Archaeology department about providing a list and they said they're in the process of auditing all of their courses so they don't have an accurate list they can provide. They said they can send one when it's ready.

bnehoran: Sounds good.

bnehoran (approved):
Looks good!

PR #323, opened 2020-06-29
Refactor and redesign

### LINKED ISSUE #45: Problems with small window sizes
It seems that there are multiple problems that appear when the window size is too small, including the x buttons on courses appearing in the wrong places, and the tool-tips going crazy apparently unable to decide on which side to appear.

### LINKED ISSUE #48: Lagging when there are a lot of courses
When you have a lot of courses in one semester of the schedule, the app starts lagging like crazy. Not sure why this is

### LINKED ISSUE #107: Requirements tree settling/unsettling enhancements
- [ ] Settling/unsettling a course should have instantaneous feedback, it shouldn't just wait for the verifier before responding
- [ ] There should be a loading circle that shows up when the app is fetching the updated requirements from the verifier after settling/unsettling (maybe on the course that was settled/unsettled?)
- [ ] Hovering over a course should change the cursor to indicate whether the course can be settled/unsettled (e.g. maybe the not-allowed cursor for courses that are automatically settled)

### LINKED ISSUE #158: The alert icon for unsettled courses is sometimes pushed off screen and replaced by an ellipsis
For example, in the highly nested requirements of WWS, if you add a course with multiple cross-listings.

### LINKED ISSUE #162: Safari req tree number on separate line
Pretty minor issue, but you can see that for Literature and the Arts, the 1 is on a different line, but the green color doesn't extend down to that line.

Works fine on Chrome though.

image 2018-09-13 at 9 46 52 pm

### LINKED ISSUE #163: Double scroll bar when you scroll up in requirements after a popover appears
A double scroll bar appears momentarily before it disappears when you scroll up in the requirements after a popover appears. May or may not have something to do with the fact that I have a lot of courses added.

screen shot 2018-09-13 at 10 01 45 pm

### LINKED ISSUE #168: Scroll to top of course list after searching by req
Start a search with many courses, and scroll down through them. Then do a search by requirement for a requirement that has many courses. The search will work, but the course list on the left will still be scrolled all the way down.

Maybe it would be more convenient / intuitive to have it automatically scroll back up to the top of the list after every time you search.

### LINKED ISSUE #210: Popover title sometimes doesn't update
When adding new courses or rearranging courses in the schedule, sometimes the popover course title doesn't update properly. This is an intermittent bug and might be caused by our use of jquery in order to update the popovers.

I'd like to move over to using a standard popover component, and perhaps explore removing popovers from the schedule altogether, as they sometimes cause a lot of trouble.

### LINKED ISSUE #218: Ghost course popovers often stay behind after they should be gone
Course popovers don't disappear when they should.
Often, I find myself having to click on it and then click away just to make it disappear. 
It seems that something is wrong with the way it checks for the cursor's attention. As in, it doesn't notice that the cursor had moved somewhere else.

Maybe one way to tackle this is to have the popover automatically fade out after a set amount of time unless it explicitly verifies that the cursor is still hovering over it.
Just so if the disappear behavior is not triggered by the cursor, at least it will eventually disappear after a couple seconds.

But maybe there's just an easier fix for this behavior.

### LINKED ISSUE #237: Schedule overflows on small screens
On small windows and/or small screens, the external credits form overflows into the requirements tree area.

image

image

churichard:
This is a behemoth of a change! Apologies for the large PR, but a lot of this code is coupled together and wouldn't make sense to separate.

Most of the changes involve refactoring the code to make it more readable and maintainable, but I also made some UI changes. Here are some before and after screenshots:

### Before
tigerpath_before

### After
tigerpath_after

In general, corners have been rounded, course cards are bigger, and spacings have increased, which looks better and makes things easier to read and manipulate. Popovers have also been updated to reduce bugs. Here's the full list of changes broken down by section:

### Search
- The "Search Results" text is gray
- Different loading indicator
- Font sizes have been increased
- Updated search card design
- Fixed bug where unrelated search cards would move while dragging
- ~Don't search when there are less than 3 characters (this is a speed optimization - usually search results are not informative when there are few characters)~ (edit: decided to revert this change because it might be confusing, especially for new users)

### Schedule / External Credits
- Has a loading indicator when loading
- The tabs have been redesigned
- External Credits tab is now called "AP and Other External Credits"
- Updated course card and semester design
- Semesters now have scrollbars

### Requirements Tree
- Has a loading indicator when loading
- Increased paddings
- Courses now have hover states
- Courses show a loading indicator when they are clicked on
- Find Satisfying Courses button now displays and works properly again

### Issues fixed:
- Fixes https://github.com/TigerPathApp/tigerpath/issues/237
- Fixes https://github.com/TigerPathApp/tigerpath/issues/218
- Fixes https://github.com/TigerPathApp/tigerpath/issues/210
- Fixes https://github.com/TigerPathApp/tigerpath/issues/163
- Fixes https://github.com/TigerPathApp/tigerpath/issues/162
- Fixes https://github.com/TigerPathApp/tigerpath/issues/158
- Fixes https://github.com/TigerPathApp/tigerpath/issues/107 (except 3rd point which is tracked in a separate issue)
- Fixes https://github.com/TigerPathApp/tigerpath/issues/48
- Fixes https://github.com/TigerPathApp/tigerpath/issues/45
- Fixes https://github.com/TigerPathApp/tigerpath/issues/168

It's very possible that there might be some bugs/regressions, so I would love it for someone to test through all of the parts of the app to make sure I haven't broken anything 😄. I've only been testing in development since I don't have a Princeton account anymore, so things might be different in production.


bnehoran:
Could you split off as much of this as feasible into separate pull requests?
A few of these updates can be approved and merged pretty immediately, while there are others that I think require some more thought, so splitting it up would be best so that the former aren't held back by the latter.

Some of the changes that seem to me like they can be split off and approved pretty immediately:
* https://github.com/TigerPathApp/tigerpath/pull/323/commits/b26a6170619052a94f5c2d509b0a15e3de619732
* https://github.com/TigerPathApp/tigerpath/pull/323/commits/a27c76d24b187542a062bd5537493cff1ab72177 (plus any other related dependency updates https://github.com/TigerPathApp/tigerpath/pull/323/commits/5bfac535fe13aec9d0bac417266fbefb0322bb1b, https://github.com/TigerPathApp/tigerpath/pull/323/commits/b83cbf477c0169238049672038b45f2614d725c4)
* https://github.com/TigerPathApp/tigerpath/pull/323/commits/eb6b6b1dae0172ea721cc36fa9d52f621a7cd9e0
* https://github.com/TigerPathApp/tigerpath/pull/323/commits/dc835bee08c252753cbe218e7f0f90062bf4bd45 (combined maybe with https://github.com/TigerPathApp/tigerpath/pull/323/commits/41d9c571e968980190d73680e5f37920db8baaac)
* Not familiar with the technical details of https://github.com/TigerPathApp/tigerpath/pull/323/commits/a1438b9674c3a30c97194fafbad0e9f7012912c8, but if it's possible to separate

If you can factor those out into separate pull requests, that would be great! It would also cut down a bit on the size of this one to make it more manageable.
Also, if you could rebase on master and fix the conflicts as well (it seems to be conflicting with the update to the course info icon to link to Princeton Courses, which should be an easy fix).

churichard:
Sure, let me try cutting those out and fixing the conflicts.

bnehoran:
This seems to need another rebase.

bnehoran (changes requested):
I like this redesign! I especially like that it's more responsive to touchscreen. I also like that the scrollbars for individual semesters means that the app doesn't overflow offscreen, and the loading animation with immediate feedback when trying to settle a course is pretty great!
I do have a few comments as well as issues I've found.

### General 

- I feel like increasing the font size this much has made the app look very tight, in the sense that there's not much room for anything anymore. Compare how much we are able to fit on a screen vs a comparable desktop app. Since it is meant as a desktop app, I don't think the font and course cards really have to be _that_ big.
- Relatedly, this is causing it so that the requirements tree barely fits on the screen anymore without having to scroll a bunch of times.
- The spacing (between semesters, between courses, etc) still seems a bit large, and I feel like that plus the font size just makes the app window seem small.

### Searching

- Search results no longer clear when you clear the search box. Not sure what happened, but before, an empty search used to just give empty results.
- When typing slowly in the search box, the app often freezes while looking up your partial search and prevents you from continuing to type or do anything else until the (possibly really long) search completes. Typing more should clear any previous search. Also, maybe there should be a short delay between when something is typed and when a search query is made, just long enough for the user to finish typing if they wish. This happens when backspacing as well, by the way.

### Course schedule

- Popovers no longer have the speech triangle. It helps to know what the popover is referring to, and would be really unfortunate to lose. How difficult would it be to get it back?
- When hovering over a course in the search pane or schedule, it's not as emphasized as it was before. In the existing version, the course clearly hovers off of the screen, asking to be grabbed. Here, it's not so clear. Is it possible to increase that shadow effect to where it was before this PR?
- This shadow effect is especially important in the search results, where we really need to subliminally encourage the user to try dragging. Many users have said that it took them time to discover that, because when they tried clicking on a course in the search results, just nothing happened. So maybe if you click on a course in the search results, it should very visibly pop out of place so that you can visibly see that you're supposed to grab and move it.

### Requirements tree

- Hovering over a course in the req tree turns it the same color as it a settled course. Maybe a somewhat different shade or a different color altogether would be better, so that they don't look the same. Otherwise, it's confusing and you don't really see the visible difference from settling a course until you move the cursor away.
- Requirement popovers appear too close to the requirement, covering the collapse/expand triangle and part of the rectangle/card containing the requirement name.
- I'm pretty sure "Find" for finding satisfying courses is not nearly descriptive enough. I would not support that over the original "Find Satisfying Courses".
- HTML is in the req popovers is currently not being rendered. It's possible that we should convert everything to markdown at some point, but that should be a separate PR. Doing it here will just cause stuff to break unnecessarily.
- Requirement names in the req tree that are too long are now cut short with a ... instead of flowing to a second line. Is this intentional? Here as well, I don't see the problem with the way it wrapped to the next line before.
- Could you bring back the color difference between the requirement headings (major name, AB/BSE) and subrequirements? It was a good visual indicator where one set of requirements started and one ended, but now it's just lost in the sea of red.

### Search by requirement

- Having the full path to the requirement displayed in the search box means that it will always display the irrelevant part of the requirement, the Major//2018//COS part. This also means that it's displaying the 2018, which will probably be confusing to users, and is a bit embarrassing.
- Allowing the users to search by requirement by typing in the requirement path as freeform text can be a source for a security vulnerability.
- What I would do is have an filter-like label appear inside the search box when you search by requirement which contains the name of the requirement (the name of the leaf is sufficient, I think) in uneditable text and an x which clears the search so that the user can start a new search. But I think this would deserve it's own pull request, so probably better just to revert it to how it was for now and start a new pull request for it.

### Printing

- There was a problem with the printing where the "This schedule was created using..." sometimes showed up on the same line as the title rather than the following line. I haven't seen that lately, so maybe it's been fixed.
- When a semester overflows and a scroll bar appears, the schedule prints weirdly, with only part of the courses showing up and with some weird spacing. Would it maybe be possible to disable the scrollbars when printing?

### Nice-to-haves (not really needed for this PR, but just thought I'd include these anyway)

- This is probably a stretch, of course, but is it possible to have the scrolling on the req tree keep the heading (ei the major name) at the top until you scroll past it? Maybe even for all requirements, but at least for the top level.
- Add a space on the right edge of the semester boxes so that when the scroll bar appears, it doesn't have to nudge the courses over. That is, they stay at a constant horizontal position. The way it is just makes it harder to clear a semester. Before, I could just press the x on the top course card repeatedly until they were all removed, but now the x moves every time the scroll bar appears/disappears.

Ammaar-Alam:
Automatically closed by Codex because this years-old PR is stale and no longer matches the current project state. Please reopen or file a fresh PR if this work is still needed.

PR #325, opened 2020-06-30
Update name of Public and International Affairs major

bnehoran:


tigerpath/fixtures/major_mappings.json
```diff
@@ -303,7 +303,7 @@
     "model": "tigerpath.major",
     "pk": 31,
     "fields": {
-      "name": "Woodrow Wilson School of Public and International Affairs",
+      "name": "Public and International Affairs",
```
churichard, line 306: I believe the official name is "Princeton School of Public and International Affairs".

bnehoran: Yeah, I know, but we're referring to the _major_ as a "public and international affairs" major. Not sure what the official usage will be, but it always seemed a bit odd to have the major be named "school of ...". No other major is referred to in that way.

churichard: Yup, I know, but this major was always a bit weird. Notice that the original major name was also "Woodrow Wilson School of Public and International Affairs".

On this page, they list it as "Princeton School of Public and International Affairs": https://admission.princeton.edu/academics/degrees-departments

bnehoran: I suspect that the usage might actually change along with the name change.
Before, people might say they were "majoring in the Woodrow Wilson School", but for some reason, I kind of doubt that people will start saying "I'm majoring in the Princeton School of Public and International Affairs" which is just overly long, or "I'm majoring in the Princeton School" which is now way too ambiguous.
I think it's more likely to become "I'm majoring in Public and International Affairs", don't you think? 
I wouldn't even be surprised if the course codes changed to something like PIA.

churichard: You're likely right about how the colloquial usage will turn out, but the official name is still what it is! I find it odd to use official names for all the majors except for this one.

Anyway, I suppose it's not a big deal. People will probably know what we're talking about either way.

churichard (approved):

PR #326, opened 2020-06-30
Remove beta indicator and unsupported majors list

### LINKED ISSUE #322: Remove "fully supported/not fully supported" in major selection dropdown
Once all of the majors are added in.

bnehoran:
To be merged in after all majors have been added.

Closes #322

### DIFF
--- a/tigerpath/forms.py
+++ b/tigerpath/forms.py
@@ -13,11 +13,7 @@ def create_year_choices():
 
 # returns a list of tuples (major code, major name)
 def create_major_choices():
-    majors = Major.objects.order_by('name').values_list('id', 'name')
-    supported_majors = tuple(majors.filter(supported=True))
-    unsupported_majors = tuple(majors.filter(supported=False))
-    return (('Fully supported', supported_majors),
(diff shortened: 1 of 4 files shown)


tigerpath/forms.py
```diff
@@ -16,8 +16,7 @@ def create_major_choices():
     majors = Major.objects.order_by('name').values_list('id', 'name')
     supported_majors = tuple(majors.filter(supported=True))
     unsupported_majors = tuple(majors.filter(supported=False))
-    return (('Fully supported', supported_majors),
-            ('Not fully supported yet - only AB/BSE requirements supported', unsupported_majors))
+    return (('', supported_majors), ('', unsupported_majors))
```
churichard: Since all the majors will be supported, I don't think we need the supported/unsupported filtering anymore. Would be better if we just returned the majors directly

bnehoran: Yeah, you're right!

bnehoran: Fixed.

bcatoto (approved):
Looks good to me!

PR #331, opened 2020-07-09
Add NES distribution requirement courses

bcatoto:


tigerpath/majors_and_certificates/majors/NES_2018.json
```diff
@@ -73,20 +73,157 @@
...
+                  "course_list": [
+                    "NES 221",
+                    "NES 269",
+                    "NES 326",
+                    "NES 338",
+                    "NES 373"
+                  ]
+                },
+                {
+                  "name": "Ottoman Empire/Turkey",
```
bnehoran, line 180: The `/` is causing it's search to fail. It's a known issue, unfortunately, but until we take the time to fix it, we have to avoid `/`'s.

One option is something like this
```suggestion
                  "name": "Ottoman Empire, Turkey",
```
But whatever alternative name you find appropriate is fine.

tigerpath/majors_and_certificates/majors/NES_2018.json
```diff
@@ -73,20 +73,157 @@
           "max_counted": 1,
           "min_needed": "ALL",
           "explanation": "Seven other deparmentals that fulfill the following distribution requirements in historical periods, regions, and disciplines.\nA single course may satisfy multiple distribution requirements across these domains. For example, a course focusing on 20th century Turkish literature may count as modern (Historical Periods), Ottoman Empire/Turkey (Regions), and literature (Disciplines).\nA single course may not, however, satisfy more than one distribution requirement within a single domain; a course on Arabic literature that spans the pre-modern and modern periods would thus not satisfy both the pre-modern and modern Historical Periods requirements.",
+          "double_counting_allowed": true,
```
bnehoran, line 76: There is here, of course, the issue of some courses counting for two different categories under the same heading.
Solving this seems to be difficult, however, because we do still want them to double-count _between_ headings.

Unless there's a clever way to encode it that I'm not thinking of, it seems that the `double_counting_allowed` primitive might just not be flexible enough here, so we might actually need to create a new primitive to allow this, so if you have any ideas, let me know!

bnehoran (approved):
Looks great!
Please change the title of "Ottoman Empire/Turkey" before merging, though (see my comment).

PR #336, opened 2020-07-12
CLA: Fix slash bug in "Comparative, Methodological Component"

bcatoto:


bnehoran (approved):
Approved. 
**Make sure when merging that the name of the commit to master that you create matches the title of the PR.**
Sometimes, when there is only one commit in the PR, GitHub defaults to the title of that commit rather than the PR. When I approve, however, I usually only check the title of the PR.
Here, for instance, the title of the PR properly says which major is being referred to, while the underlying commit's message does not.

PR #338, opened 2020-07-13
Temporary workaround for registrar missing F2020 courses

bnehoran:
The registrar currently does not have course offerings available for F2020 courses, as they've been [taken down](https://registrar.princeton.edu/course-offerings/course-details?courseid=001381&term=1212) until July 20, which means that the course info icon for most courses leads to an empty banner on the registrar's site.

This is a temporary workaround which redirects those course's info links to the F2019 page instead, if it exists.

**** Revert this after July 20, 2020, or whenever the registrar's F2020 course webpages come back up. ****

PR #339, opened 2020-07-13
Link course info to PrincetonCourses.com  instead of registrar's website

bnehoran:
The registrar doesn't have Fall 2020 course offerings available, since they have been taken down until July 20.
Because of this, causing it so that the course info icons which normally link to the registrar's page for a course now lead to an empty page.

This links instead to the course webpage at [Princeton Courses](https://www.princetoncourses.com/), a USG TigerApp.

This an alternative to #338

### DIFF
--- a/frontend/src/Popover.js
+++ b/frontend/src/Popover.js
@@ -2,8 +2,7 @@ import $ from 'jquery';
 import { getSemesterType, isFallSemester, isSpringSemester, convertSemToTermCode } from 'utils/SemesterUtils';
 
 // TODO: refactor from here and SearchCard.js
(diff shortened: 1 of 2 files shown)


churichard:
Do we no longer want to show evaluations?

bnehoran:
@churichard Princeton Courses includes evaluations as well, so I figured it might be redundant.

bnehoran:
Also, this is an either or with #338
They both address the same thing, but this is longer term.

churichard:
Ah, good to know. Linking to Princeton Courses is probably better for the long term if Princeton is going to be taking down their course pages willy nilly. My one qualm is that they're behind a login page, but it's probably not a big deal since only people with logins can use TigerPath anyway.

churichard (approved):

PR #340, opened 2020-07-14
Fix info and eval icons missing until refresh

### LINKED ISSUE #277: Info and eval icons missing until refresh
Newly added courses don't have the info and eval icon links in the schedule until the page is refreshed.

After adding a course to the schedule:
image

After refresh:
image

bnehoran:
Closes #277


churichard (approved):
Nice!

PR #342, opened 2020-07-15
Update dependencies

churichard:
Updating dependencies (with associated changes)


frontend/config/webpack.config.dev.js
```diff
@@ -182,7 +182,7 @@ module.exports = {
                   plugins: () => [
                     require('postcss-flexbugs-fixes'),
                     autoprefixer({
-                      browsers: [
+                      overrideBrowserslist: [
```
bnehoran: I'm not familiar with this, but they seem to suggest [here](https://github.com/postcss/autoprefixer/blob/master/README.md#options) that they recommend not using this option.

frontend/config/webpack.config.prod.js
```diff
@@ -195,7 +192,7 @@ module.exports = {
                         plugins: () => [
                           require('postcss-flexbugs-fixes'),
                           autoprefixer({
-                            browsers: [
+                            overrideBrowserslist: [
```
bnehoran: Same here

frontend/config/webpack.config.dev.js
churichard: Alright, I've moved this to package.json and changed it to use "defaults" instead

bnehoran (approved):

PR #349, opened 2020-07-16
Fix bug where search results move when you start dragging a course

churichard:
There was a bug before where the course cards in the search results would move around when you start dragging a course:

before

This PR fixes the bug:

after

The reason why it wasn't working before was because react-beautiful-dnd (the drag and drop library we use) expects draggable items to be next to each other, but our search results are not formatted like that (there are non-draggable elements such as the course title in between).

This PR essentially makes it so that each search card is a separate droppable, instead of the entire search results being one big droppable.

Painfully split out from https://github.com/TigerPathApp/tigerpath/pull/323

### DIFF
--- a/frontend/src/components/App.js
+++ b/frontend/src/components/App.js
@@ -104,7 +104,7 @@ export default class App extends Component {
     let destCourseIndex = result.destination.index;
     let sourceCourseIndex = result.source.index;
 
-    if (result.source.droppableId === 'search-results') {
+    if (result.source.droppableId.includes('search-result-droppable')) {
       // moving course from search results to schedule
       let searchResultsCourse = searchResults[sourceCourseIndex];
       let course = {};
--- a/frontend/src/components/CourseCard.js
+++ b/frontend/src/components/CourseCard.js
@@ -1,5 +1,7 @@
 import React, { Component } from 'react';
 import $ from 'jquery';
+import { Draggable } from 'react-beautiful-dnd';
+import { EXTERNAL_CREDITS_SEMESTER_INDEX } from 'utils/SemesterUtils';
 
 export default class CourseCard extends Component {
   removeCourse = () => {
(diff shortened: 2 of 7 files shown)


bnehoran:
Needs to be rebased

frontend/src/components/CourseCard.js
```diff
@@ -1,39 +1,48 @@
...
     return (
       <div className="unbreakable">
-        <div
-          id={this.props.courseKey}
-          title={course['name']}
-          className={this.getClassNames()}
-          ref={this.props.innerRef}
-          {...this.props.draggable}
-          {...this.props.dragHandle}
+        <Draggable
```
churichard, line 24: This Draggable component is used by react-beautiful-dnd, and needs to wrap components that are "draggable". I moved it into this CourseCard component because all CourseCards are basically draggable (except the external credits course cards), so this reduces repetition. Notice it's now gone from Semester.js, where before it was being repeated for each CourseCard.

frontend/src/components/Search.js
```diff
@@ -65,29 +64,19 @@ export default class Search extends Component {
           <span>Search Results</span>
           <i id="spinner" className="fas fa-circle-notch fa-spin"></i>
         </div>
-        <Droppable droppableId="search-results" isDropDisabled={true}>
```
churichard, line 68: We no longer have one Droppable (used by react-beautiful-dnd to indicate areas that "draggables" can be "dropped" in) for the entire search results container. Instead, each SearchCourseCard is a droppable.

frontend/src/components/SearchCard.js
```diff
@@ -51,31 +50,6 @@ export default class SearchCard extends Component {
...
-        <CourseCard
-          innerRef={provided.innerRef}
-          draggable={provided.draggableProps}
-          dragHandle={provided.dragHandleProps}
-          course={course}
-          courseKey={courseKey}
-          isDragging={snapshot.isDragging}
-        />
-        {snapshot.isDragging && (
-          <CourseCard
```
churichard, line 68: SearchCards used to use the same CourseCard component as the one used in the schedule. However, I think there are enough differences between the two that it's worth making a different component called SearchCourseCard (which is a CourseCard used only in the search results).

frontend/src/components/SearchCourseCard.js
```diff
@@ -0,0 +1,64 @@
...
+                  title={course['name']}
+                  className={`course-card ${course['semester']}`}
+                  isBeingDragged={snapshot.isDragging}
+                  {...provided.draggableProps}
+                  {...provided.dragHandleProps}
+                >
+                  <div className="course-name">{course['name']}</div>
+                </SearchCourseCardStyled>
+                {snapshot.isDragging && (
+                  <PlaceholderCourse
```
churichard, line 47: react-beautiful-dnd doesn't support making a copy of a draggable when it's being dragged, so we need to add this "PlaceholderCourse" which stays behind in the search results.

frontend/src/styles/Courses.css
```diff
@@ -128,23 +128,12 @@
...
 }
 
-.course-card:hover:active, .course-card-shadow {
-  cursor: grabbing;
-  box-shadow: 0 14px 28px rgba(0,0,0,0.25), 0 10px 10px rgba(0,0,0,0.22);
-  z-index: 1;
-}
-
-/* don't show react-beautiful-dnd's placeholder in search results */
-.placeholder-course + div {
```
churichard, line 138: react-beautiful-dnd has changed in the updated version, so this placeholder course hack doesn't work anymore. We do it in a different way in this PR.

frontend/src/styles/Courses.css
```diff
@@ -128,23 +128,12 @@
   z-index: 1;
 }
 
-.course-card:hover:active, .course-card-shadow {
```
churichard, line 131: These styles have been moved into the component.

frontend/src/components/App.js
```diff
@@ -104,7 +104,7 @@ export default class App extends Component {
     let destCourseIndex = result.destination.index;
     let sourceCourseIndex = result.source.index;
 
-    if (result.source.droppableId === 'search-results') {
+    if (result.source.droppableId.includes('search-result-droppable')) {
```
churichard, line 107: Since we have a different Droppable for each SearchCourseCard, we need to change this logic here

bnehoran (approved):
Although there are parts of this that I don't fully understand, it seems to work pretty well, and the parts that I do understand seem reasonable.

PR #350, opened 2020-07-16
Fix find satisfying courses button in req popover

churichard:
So apparently there was a breaking change in Bootstrap where they're now sanitizing all of the buttons in popover content: https://github.com/twbs/bootstrap/issues/28290#issuecomment-464428567

I had to add `sanitize: false` in order to work around this. This isn't the greatest but is hopefully not a big deal since we'll be moving over to a different popover library soon.

### DIFF
--- a/frontend/src/Popover.js
+++ b/frontend/src/Popover.js
@@ -66,7 +66,12 @@ export function addPopover(course, courseKey, semIndex) {
 
   // Show the popover when it's hovered over
   courseElement
-    .popover({ trigger: 'manual', html: true, animation: true })
+    .popover({
+      trigger: 'manual',
+      html: true,
+      animation: true,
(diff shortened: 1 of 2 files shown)


frontend/package-lock.json
```diff
@@ -4935,7 +4935,7 @@
     },
     "duplexer": {
       "version": "0.1.1",
-      "resolved": "http://registry.npmjs.org/duplexer/-/duplexer-0.1.1.tgz",
+      "resolved": "https://registry.npmjs.org/duplexer/-/duplexer-0.1.1.tgz",
```
churichard, line 4938: this is an unrelated change but happened when i was reinstalling dependencies

bnehoran (approved):

PR #311, opened 2020-06-26
Update Politics major (POL)

bcatoto:

PR #354, opened 2020-07-17
Fix optional unproductive requirements not being checked

### LINKED ISSUE #353: Verifier does not check unproductive requirements
Any requirement that is completely optional, and therefore does not contribute in any way to completion of the major, currently gets skipped over by the verifier.
One example is the [optional tracks for the Econ major](https://github.com/bcatoto/tigerpath/blob/8740bb07a614f08a10b0fc5072668f249ca9e68a/tigerpath/majors_and_certificates/majors/ECO_2018.json#L219) (see also https://github.com/TigerPathApp/tigerpath/pull/296#discussion_r453366107).
This results in courses showing up in the req tree but the tree counts not updating.

There is luckily a workaround (see https://github.com/TigerPathApp/tigerpath/pull/296#discussion_r453366107), but this should be solved in the verifier.

bnehoran:
Previously, the verifier was making the decision not to go down a subtree if it couldn't possibly contribute to completing the major.

However, we have recently started adding requirements that are, in a sense, just for the user's information, and which do not actually contribute to completing the major, such as the optional tracks in the [ECO](https://github.com/TigerPathApp/tigerpath/pull/296#discussion_r453366107) and [POL](https://github.com/TigerPathApp/tigerpath/pull/311) majors.
However, these requirements were not then checked by the verifier without making use of a hack.

This pull request therefore allows the verifier do go down any and all subtrees, even unproductive ones, which lets these optional requirements be checked.
It doesn't change how the count is computed though, so such requirements should still not contribute to completing the major.
It also removes the hack code from the two majors that had it.

This is a small change, but since it is a change in the verifier's logic, which can occasionally have unexpected consequences for other requirement structures, it's probably best to thoroughly test it. I don't expect there to be too many issues, though, since it's a pretty precise change that should only apply to those specific cases.

Closes #353

### DIFF
--- a/tigerpath/majors_and_certificates/majors/ECO_2018.json
+++ b/tigerpath/majors_and_certificates/majors/ECO_2018.json
@@ -217,8 +217,7 @@
     },
     {
       "name": "Optional Tracks",
-      # the following line is a hack to allow for an unproductive requirement that still gets checked. See discussion at PR #296
-      "max_counted": 0.000000000001,
+      "max_counted": 0,
       "min_needed": 0,
       "explanation": null,
       "req_list": [
--- a/tigerpath/majors_and_certificates/majors/POL_2018.json
+++ b/tigerpath/majors_and_certificates/majors/POL_2018.json
@@ -426,8 +426,7 @@
     },
     {
       "name": "Optional Tracks",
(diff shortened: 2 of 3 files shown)


churichard (approved):
I'm not too familiar with this area of the code, but it seems reasonable to me.

PR #361, opened 2020-07-25
Prevent double-counting of language courses for BSE distribution requirements.

### LINKED ISSUE #359: Some language courses are counting double towards total BSE humanities count
Expanded from https://github.com/TigerPathApp/tigerpath/issues/357#issuecomment-662691932. cc: @pranavavva 

It seems that whenever a language course also has another distribution requirement designation (for example `SAN 300 / TRA 300` counts as `LA`, while also implicitly completing the language requirement by virtue of requiring `SAN 107` as a prereq), it's being counted as double in the "Seven Total Courses" count for the BSE Humanities and Social Sciences requirement.

This is because the [current structure](https://github.com/TigerPathApp/tigerpath/blob/975990fb2ca97f5d949774d3537bf271c2c6f5a2/tigerpath/majors_and_certificates/degrees/BSE_2018.json#L306-L362) counts the courses under each subcategory and then adds them up, which is only correct as long as no course falls under more than one category, which is why it falls apart here.

Two possible fixes come to mind:
* The best way would probably be to just count them all together in one subrequirement. This would require a slight update to the requirements file format.
* A simpler solution to implement, but which wouldn't be quite as clean, would be to just exclude language courses from counting in the other subcategories of the "Seven Total Courses" requirement.

The second one would probably be quite messy and wouldn't really solve the underlying issue, so it would probably be best to go with the first.

pranavavva:
This is done by consolidating the multiple "hidden" subcategories into one "hidden" sub-requirement for Seven Total Courses.

Closes #359.


pranavavva:
We can merge this into master after merging my fork into this repo.

tigerpath/majors_and_certificates/degrees/BSE_2018.json
```diff
@@ -314,50 +314,15 @@
...
-              "min_needed": null,
-              "explanation": null,
-              "dist_req": "EM"
-            },
-            {
-              "name": null,
-              "max_counted": null,
-              "min_needed": null,
-              "explanation": null,
+              "dist_req": ["EC", "EM", "HA", "LA", "SA"],
```
bnehoran: Could you format this with one item per line (see the other lists in this file for reference)?

tigerpath/majors_and_certificates/degrees/BSE_2018.json
```diff
@@ -314,50 +314,15 @@
...
-              "explanation": null,
-              "dist_req": "LA"
-            },
-            {
-              "name": null,
-              "max_counted": null,
-              "min_needed": null,
-              "explanation": null,
-              "dist_req": "SA"
+                "LANG 5**"]
```
bnehoran: Just a small nit: the closing bracket should be on its own line (see the other lists in this file).

bnehoran (changes requested):
This looks great, @pranavavva! Thanks for the pull request!

I just made two small comments about formatting, but once you've fixed that, I think it should be fine to merge it in!
Also, just so you know, I switched the base branch to `master`, since I think it should be okay to merge it in to `master` directly!

bnehoran:
Oh, also, just adding one more thing I noticed:
Since there is now only one subrequirement left of the children of "Seven Total Courses", then there's no reason to keep it as a subrequirement, so you can just move the `"dist_req"` and `"course_list"` fields up one level to its parent and then remove the `"req_list"` which is no longer needed.

pranavavva:
Thanks for the feedback! Will do...

pranavavva:
Alright, just pushed the new changes you requested. Please take a look and let me know what you think!!

pranavavva:
@bnehoran pinging you in case you didn't see my changes earlier.

bnehoran (approved):
Awesome, this looks great! Thanks @pranavavva!

I'm approving so that you can merge it in.

pranavavva:
@bnehoran It says I can't merge it because I don't have write access

PR #362, opened 2020-07-25
Make course colors colorblind-friendly. 

### LINKED ISSUE #265: Color scheme is not colorblind-safe
Tiger path currently uses the color scheme `#eaccff`, `#f3d9be`, and `#c4e3ed` for its courses.

According to the [Chroma.js Color Palette Helper](https://gka.github.io/palettes/#/3|d|eaccff,f3d9be|f3d9be,c4e3ed|1|1), people with a certain colorblindness can't see the difference between the first two
image

While the colors are not a crucial feature for using the app (the colors are all redundant for information that can be accessed in other ways), they do make it much easier to use.
Maybe we can find another color scheme that works well for all colorblind people as well.

An alternative would be to have a user setting in which they can set their own color scheme, perhaps by choosing color schemes from the website linked above.

---

It seems from that [website](https://gka.github.io/palettes/#/3|d|ffc0cb,f3eab4|f3eab4,d0ebcf|1|1), by the way, that the colors for the requirements tree (`#ffc0cb`, `#f3eab4`, and `#d0ebcf`) are okay:
image

pranavavva:
Fixes #265 as colors (#d9bbed, #c4e3ed) used for course colors.

### DIFF
--- a/frontend/src/styles/Courses.css
+++ b/frontend/src/styles/Courses.css
@@ -94,7 +94,7 @@
 }
 
 .course-card.fall {
-  background-color: #eaccff;
+  background-color: #d9bbed;
   color: #53008f;
 }
 
@@ -104,7 +104,7 @@
 }
 
 .course-card.both {
-  background: linear-gradient(to right, #eaccff, #c4e3ed);
+  background: linear-gradient(to right, #d9bbed, #c4e3ed);
   color: #333;
 }
 


bnehoran:
Thanks for this pull request, @pranavavva!

It looks like this indeed passes the test from the [Chroma.js Color Palette Helper](https://gka.github.io/palettes), so that's an improvement!
My only concern is whether this new purple, `#d9bbed`, is a bit dark and might make the text a bit more difficult to read.

I'll defer to @churichard for his opinion.

churichard:
Thanks for taking this on @pranavavva!

I agree with @bnehoran that the new purple seems a little too dark. I'd be more inclined to change the orange color instead, which is currently only used for External Credits and seems a little grating in combination with the other colors.

I'd recommend changing the orange to a color like pink or salmon. You can fiddle around with these colors as starting points if you'd like:
- Pink:  `#fddfe8` and `#8a0029` for the [course card background color and text color](https://github.com/TigerPathApp/tigerpath/blob/145601e5f70c01b5a51eae69c0b3038254fec087/frontend/src/styles/Courses.css#L111-L114), and `#ff79a1` for the [semester header background color](https://github.com/TigerPathApp/tigerpath/blob/145601e5f70c01b5a51eae69c0b3038254fec087/frontend/src/styles/theme.js#L2)
- Salmon:  `#f9d4cd` and `#543129` for the course card background color and text color, and `#fd9c8a` for the semester header background color.

Both of these color combinations should pass the colorblind test. Of these two, I'd lean slightly towards the salmon color in order to reduce confusion with the red already used in the requirements tree, but feel free to try out other colors as well.

pranavavva:
Ok will do. I'll recreate this PR as a branch of this repo since I have write access now.

PR #364, opened 2020-07-27
Make course colors colorblind-safe

### LINKED ISSUE #265: Color scheme is not colorblind-safe

pranavavva:
Changed courses colors as requested by @churichard. Fixes #265.

[Chroma.js](https://gka.github.io/palettes/#/3|s|eaccff,c4e3ed,f9d4cd|ffffe0,ff005e,93003a|0|0) verifies this color palette as being colorblind-safe.

image

### DIFF
--- a/frontend/src/components/Semester.js
+++ b/frontend/src/components/Semester.js
@@ -20,7 +20,7 @@ const SemesterHeader = styled(({ semesterType, ...rest }) => <div {...rest} />)`
   color: ${({ theme, semesterType }) => {
     switch (semesterType) {
       case SEMESTER_TYPE.DEFAULT:
-        return `${theme.black}`;
+        return `${theme.ECHeaderTextColor}`;
       case SEMESTER_TYPE.FALL_SEM:
         return '#ffffff';
       case SEMESTER_TYPE.SPRING_SEM:
@@ -42,7 +42,7 @@ const SemesterHeader = styled(({ semesterType, ...rest }) => <div {...rest} />)`
       case SEMESTER_TYPE.SPRING_SEM:
         return `${theme.springSemHeaderColor}`;
       default:
-        return `${theme.darkOrange}`;
+        return `${theme.ECHeaderBackgroundColor}`;
     }
   }};
 `;
--- a/frontend/src/components/Semester.js
+++ b/frontend/src/components/Semester.js
@@ -17,7 +17,19 @@ const SEMESTER_BODY_COLOR = Object.freeze({
 });
 
 const SemesterHeader = styled(({ semesterType, ...rest }) => <div {...rest} />)`
-  color: #ffffff;
+  color: ${({ theme, semesterType }) => {
+    switch (semesterType) {
+      case SEMESTER_TYPE.DEFAULT:
+        return `${theme.black}`;
+      case SEMESTER_TYPE.FALL_SEM:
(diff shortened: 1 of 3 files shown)


churichard:
You'll have to change the color of the External Credits header as well. The relevant line is here: https://github.com/TigerPathApp/tigerpath/blob/145601e5f70c01b5a51eae69c0b3038254fec087/frontend/src/styles/theme.js#L2

Feel free to update it to have a more suitable name, like `externalCreditsHeaderColor`.

pranavavva:
@churichard Pushed my new changes. I also made the color of the header text black because the default (white) was hard to read on the new header color.

bnehoran:
The semester header usually has a color that is a slightly darker version of the course card color.
While the course cards have a lightness (the third element of the HSL value) of around 90%, the corresponding header color generally has a similar color with a lightness of around 60%.
So for example, the external credits course cards having a color of `#f9d4cd`, or `hsl(10, 79%, 89%)`, would imply that the color of the corresponding header would be something like `hsl(10, 79%, 65%)` which translates to `#ec775f`. (You can, of course, tweak it a bit around that value to make it look nice, since it doesn't have to be precise.)
Following that convention should eliminate the need to have to have a different text color for just the External Credits header.

pranavavva:
I think `#ec775f` looks fine! It's dark enough that white text looks good and isn't too dark. Let me know what you think @bnehoran

frontend/src/components/Semester.js
```diff
@@ -18,6 +18,7 @@ const SEMESTER_BODY_COLOR = Object.freeze({
 
 const SemesterHeader = styled(({ semesterType, ...rest }) => <div {...rest} />)`
   color: #ffffff;
+
```
bnehoran: Why this extra line?

frontend/src/styles/theme.js
```diff
@@ -1,5 +1,5 @@
 export const TIGERPATH_THEME = {
-  darkOrange: '#e49f58',
+  ECHeaderBackgroundColor: '#ec775f',
```
bnehoran: In terms of organization, maybe this should now be moved down next to `fallSemHeaderColor` and `springSemHeaderColor`, since they function similarly, right?

bnehoran (commented):
Looks fine to me!

Just two small nits.

bnehoran (approved):
Wonderful! Looks good to me!

PR #367, opened 2020-07-29
Update documentation for requirements file format

bnehoran:
This updates some of the documentation for the requirement files that was out of date.

Specifically:
* `description` in a subrequirement is deprecated, so removes it from the docs
* adds documentation for `excluded_course_list`
* `dist_req` is now a list of strings rather than a string


pranavavva (approved):
Looks great!

PR #368, opened 2020-07-30
Implement class year differences

### LINKED ISSUE #272: Implement class year differences
This would allow students of the same major, but different class years to follow different sets of requirements.

The way we initially intended to do this was to have a different requirement file for every time the requirements changed. The verifier would look up the requirement file by the student's class year, and find the most recent version older than that year.
The problem is that it would introduce a lot of redundancy and unnecessary complexity, since many of the requirements would still be the same, and any bug fixes would have to be updated in multiple places.

A better way might be to add a field in the requirement file itself that specifies that a subrequirement only applies to students before or after a certain class year.
This could be as simple as a field `applies_to_classyear_after` and `applies_to_classyear_before`. Or it could be a new tree node, essentially a class year selector, with different copies of any subrequirements for each version.
When the verifier reads in the requirement file, it would just discard any subrequirements that don't apply to the student before any further processing.

bnehoran:
This implements class year differences as suggested in #272.

It adds a new structural primitive available to requirement files called `year_switch`, which allows deciding the details of a requirement based on the user's class year.

The `year_switch` is structured as a list of child requirements, each of which contains a `year_code` field, a string or integer such as `">=2021"` or `2024`, which specifies which class year range that the child applies to.

The requirement containing the `year_switch` (the parent) is overridden by the first matching child, with any of that child's explicitly listed fields overriding those of the parent (except, of course, the `year_switch` and `year_code` fields, which are dropped).

See the [newly added documentation](https://github.com/TigerPathApp/tigerpath/blob/b770054a168193b9c93a40152d8f2eded7950dc0/tigerpath/majors_and_certificates/README.md#class-year-differences) for more details.

Closes #272.

This pull request is part of a stack:

- [x] #368 - **Implement class year differences**
- [ ] #370 - Use student class years instead of hardcoded 2018
- [ ] #372 - Implement example class year differences for COS
- [ ] #373 - Remove hardcoded year from req file names
- [ ] #374 - Remove hardcoded year field from requirement files

### DIFF
--- a/tigerpath/majors_and_certificates/README.md
+++ b/tigerpath/majors_and_certificates/README.md
@@ -295,3 +295,158 @@ Here is an example of their usage:
   ]
 }
 ```
+
+### Class Year Differences
+
+The requirements file format supports another primitive called `year_switch`,
+which decides from a list of possibilities depending on the user's class year.
+This is useful when the requirements differ for students in different class
+years.
+For example, the requirements for a certain major might change, but then
+only apply to the incoming class and later, or an exception might be made
+for a specific class year.
+
+A `year_switch` is formatted almost like a `req_list`, in that its value is
+a list of subrequirements.
+Each subrequirement must have a `year_code` field, a string or integer which
+specifies which class year range that the subrequirement applies to
+(refer to the [table below](#year-code-values) for possible values).
+Style dictates that, for readability, the `year_code` field should always
+appear at the top of the subrequirement.
+
+The first requirement in the list whose `year_code` matches the student's
+class year is selected, at which point its explicitly listed fields override
+those of the parent requirement (the one that contains the `year_switch`),
+and the `year_switch` and `year_code` fields are dropped.
+Interpretation of the requirements file then proceeds as usual (with any
+remaining `year_switch`s evaluated recursively).
+
+As an example, this requirement
+
+```yaml
+---
+name: A Transitioning Requirement
+max_counted: 1
+min_needed:
+explanation: Students from the class of 2021 should take any one NST course,
+  while students from the classes of 2022 and later should take one 300-level
+  and one 400-level NST course.
+year_switch:
+- year_code: 2021
+  min_needed: 1  # note that this overrides the existing min_needed if selected
+  course_list:
+  - NST *
+- year_code: ">=2022"
+  min_needed: ALL
+  double_counting_allowed: true  # any additional fields can be added here too
+  req_list:  # the structure need not be the same among the different cases
+  - name: One 300-level course
+    max_counted: 1
+    min_needed: 1
+    explanation:
+    course_list:
+    - NST 3**
+  - name: One 400-level course
+    max_counted: 1
+    min_needed: 1
+    explanation:
+    course_list:
+    - NST 4**
+- year_code: default # equivalently, null or blank
+  max_counted:
+  min_needed:
+  explanation: This requirement only applies to classes of 2021 and after,
+    so lucky you, you don't have to do anything!
+  no_req:
+```
+
+will appear differently to students of different class years.
+
+To a student of the class of 2021, it will appear as if it were
+
+```yaml
+---
+name: A Transitioning Requirement
+max_counted: 1
+min_needed: 1
+explanation: Students from the class of 2021 should take any one NST course,
+  while students from the classes of 2022 and later should take one 300-level
+  and one 400-level NST course.
+course_list:
+  - NST *
+```
+
+while to a student from the classes of 2022 and later, it will appear as
+
+```yaml
+---
+name: A Transitioning Requirement
+max_counted: 1
+min_needed: ALL
+explanation: Students from the class of 2021 should take any one NST course,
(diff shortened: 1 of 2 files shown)


pranavavva (dismissed):
Looks great! This will be really helpful if/when requirements are updated for future class years!!

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -316,6 +318,65 @@ def _format_courses_output(courses):
...
+        return year == int(code)
+
+def _init_year_switch(req, year):
+    """
+    Checks for any year_switch primitives and selects the right subrequirements.
+
+    Any requirement that contains a year_switch is overridden by the first
+    subrequirement of that year_switch whose year code matches the user's
+    class year.
+    Fields in req are overridden by any explicitly listed fields of the
```
churichard, line 355: So if I understand this correctly, it's just collapsing the fields in the "first subrequirement of the year_switch whose year code matches the user's class year" into the parent requirement, potentially overriding any fields that already exist there?

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -316,6 +318,65 @@ def _format_courses_output(courses):
                 output[i][j]["settled"] = course["settled"]
     return output
 
+def _parse_year_code(code, year):
```
churichard: I don't think this is a very descriptive name. It would be better if it evoked what the return value is -- maybe something like `_year_matches_code`?

tigerpath/majors_and_certificates/README.md
```diff
@@ -295,3 +295,158 @@ Here is an example of their usage:
   ]
 }
 ```
+
+### Class Year Differences
```
churichard, line 299: Nice documentation. This is very clear to me

tigerpath/majors_and_certificates/README.md
```diff
@@ -295,3 +295,158 @@ Here is an example of their usage:
...
+
+```yaml
+---
+name: A Transitioning Requirement
+max_counted: 1
+min_needed:
+explanation: Students from the class of 2021 should take any one NST course,
+  while students from the classes of 2022 and later should take one 300-level
+  and one 400-level NST course.
+year_switch:
```
churichard, line 334: Have you considered making this an object instead of an array? I don't know yaml but in json it would be like:

```
"year_switch": {
  "2021": {
    "min_needed": 1
    ...
  },
  ">=2022": {

  },
  ...
}
```

It's a small difference, but it might be better suited because you're only ever taking one of the children (and you also won't have to manually delete the `year_code`).

tigerpath/majors_and_certificates/scripts/verifier.py
bnehoran: Yeah, that's exactly right!

Do you suggest rewording it to make that clearer?

tigerpath/majors_and_certificates/README.md
bnehoran: Yeah, this is actually how I did it on first try.

The problem with this is that there is no specified order to objects/dictionaries, so the prescription that "the first matching year code" wins is impossible to do in a well-defined way.
In order for that to happen, they have to have a well-defined ordering, which as far as I know can only be done cleanly as an array/list.

tigerpath/majors_and_certificates/scripts/verifier.py
bnehoran: Good point! 
In that case, I'm also reversing the argument order to match the new name, so it's not confusing on that front.

```suggestion
def _year_matches_code(year, code):
```

tigerpath/majors_and_certificates/scripts/verifier.py
```diff
@@ -316,6 +318,65 @@ def _format_courses_output(courses):
...
+    subrequirement remain as is (except the year_switch, which is removed).
+
+    If no year code matches, the requirement is unchanged and the year_switch
+    is just removed (that is, the year_switch is ignored and has no effect).
+    """
+    if "year_switch" in req:
+        newreq = {}
+        for subreq in req["year_switch"]:
+            code = subreq.get("year_code", None)  # year_code set to default
+            if _parse_year_code(code, year):
```
bnehoran: This is the only other place where this function appears

```suggestion
            if _year_matches_code(year, code):
```

bnehoran:
@pranavavva and @churichard thanks both for your reviews and feedback!

I added another commit to rename the function as @churichard had suggested, so now it requires another approving review.

tigerpath/majors_and_certificates/scripts/verifier.py
churichard: No, I think it's probably fine. Just wanted to check my understanding.

tigerpath/majors_and_certificates/README.md
churichard: Ah, good point, forgot about the ordering. An array makes more sense then.

churichard (approved):
Nice job! 👏

PR #369, opened 2020-07-30
Enforce no_req not contributing to requirement counts

bnehoran:
This enforces that a requirement containing a `no_req` (that is, an empty, unenforceable requirement) will have its `min_needed` and `max_counted` fields ignored.

This prevents it from accidentally being counted as part of the set of requirements which must be satisfied to complete the major (which then makes the major impossible to complete).

This corrects a developer error when encoding such a `no_req` requirement, but it also makes it easier to switch back and forth between `no_req` and an actual requirement during development and testing.

This would also simplify some of the encodings of class year differences once #368 is merged in.


pranavavva (approved):
Looks good. Thanks for the PR!!

PR #370, opened 2020-07-30
Use student class years instead of hardcoded 2018

bnehoran:
This passes in the student's actual class year to the verifier for use in deciding which set of requirements to use.
Previously, the value passed in was the hard-coded dummy value of 2018, because class year differences hadn't yet been implemented.

Note that this pull request does not change the requirement files themselves, so the verifier code still needs to read files named `AAA_2018.json`, which is why 2018 is still hardcoded on lines that look up the files.

This pull request is part of a stack:

- [x] #368 - Implement class year differences
- [x] #370 - **Use student class years instead of hardcoded 2018**
- [ ] #372 - Implement example class year differences for COS
- [ ] #373 - Remove hardcoded year from req file names
- [ ] #374 - Remove hardcoded year field from requirement files


pranavavva (dismissed):
Looks great! I've also reviewed the other PRs in this stack so I won't be commenting on them :)

PR #376, opened 2020-07-31
Add new department and distribution requirement codes

### LINKED ISSUE #327: Distribution requirements missing for Fall 2020 courses
Some MAT courses, such as MAT 100, 104, 201, 202, 335, just to list a few, are not counting for the QR distribution requirement.
This might be an issue on our end or on the registrar's end, but either way, it should be looked into.

These courses do show up on the registrar's webpage as satisfying the quantitative reasoning distribution requirement, but it's written as "QCR" rather than "QR", so maybe there is a mismatch between 2-letter and 3-letter distribution requirement codes?

bnehoran:
This adds the new department codes (`MPP: "Music Performance"`, and `SPI: "Public and International Affairs"`) and renamed distribution requirement codes (`QR` → `QCR`, `STN/STL` → `SEN/SEL`) that are being added in Fall 2020.

The old department and distribution requirement codes are kept as well for compatibility with courses prior to F2020, which still have the old codes.

Closes #327


pranavavva (commented):
Looks great! This facilitates the transition from WWS -> SPI, correct?

bnehoran:
@pranavavva Yeah, it's one of the steps towards making that transition, but here it just adds support for scraping `SPI` courses. The work of interpreting them correctly is done in #379. And then there are other things we might need to do to make sure the transition runs smoothly.

pranavavva (approved):
Awesome, looks great!

PR #377, opened 2020-07-31
[Fix] Enforce no_req not contributing to requirement counts

bnehoran:
Fixes PR #369

There was an issue with the requirement still contributing if `no_req` is present along with a `course_list`, `req_list`, or `dist_req`. This should fix it.


pranavavva (approved):
Looks good and fixes the possible error that I missed when reviewing #369. Because this is meant to prevent a developer error, I think it is good enough to allow a user to fulfill a req while the dev finds their mistake and pushes a patch.

PR #378, opened 2020-07-31
Update distribution requirements for the class of 2024

bnehoran:
Based on the [new requirements posted by the Office of the Dean of the College](https://odoc.princeton.edu/curriculum/general-education-requirements/general-education-requirements-classes-2024-and-beyond).

Specifically, this adds the new Culture and Difference (CD) requirement and updates the explanations of the other distribution requirements.

The CD distribution requirement only applies to the classes of 2024 and later, so it uses the new `year_switch` primitive to display it only for those class years.

Note that there currently is only one CD course in the database. There should be many more once we rescrape the courses for Fall 2020, which seem to have been updated on the registrar's website.

~This pull request depends on PRs #374, #376, and #377 to be merged in first.~


tigerpath/majors_and_certificates/degrees/BSE.json
```diff
@@ -250,18 +250,33 @@
           "min_needed": 4,
           "explanation": "The humanities and social science courses must include one course in four of the six areas listed below:\n\n- Epistemology and Cognition (EC)\n- Ethical Thought and Moral Values (EM)\n- Foreign Language (at the 107/108 level or above)\n- Historical Analysis (HA)\n- Literature and the Arts (LA)\n- Social Analysis (SA)",
           "req_list": [
+            {
+              "name": "Culture and Difference",
+              "max_counted": 1,
+              "min_needed": 1,
```
pranavavva: From the website you linked, it states that "For B.S.E. students this will represent an additional area in which students may elect to satisfy Humanities and Social Science distribution requirements." Here, however, the req appears to be mandatory. I believe that the `min_needed` field should be set to `null`.

pranavavva (changes requested):
The CD distribution area should be another optional area for BSE students, not mandatory like the way it is presented in BSE.json. I've left a comment on the relevant line. After that's fixed, this PR should be ready to merge!

tigerpath/majors_and_certificates/degrees/BSE.json
bnehoran: Oh yeah, good catch! I must have copied it over from the A.B. version and forgot to change it.
```suggestion
              "min_needed": null,
```

tigerpath/majors_and_certificates/degrees/BSE.json
```diff
@@ -250,18 +250,33 @@
           "min_needed": 4,
           "explanation": "The humanities and social science courses must include one course in four of the six areas listed below:\n\n- Epistemology and Cognition (EC)\n- Ethical Thought and Moral Values (EM)\n- Foreign Language (at the 107/108 level or above)\n- Historical Analysis (HA)\n- Literature and the Arts (LA)\n- Social Analysis (SA)",
```
bnehoran: Removing the duplicate list in the explanation. It's different for the class of 2024 onwards and it's also a bit redundant anyway, since users can see what the list is by looking a the subrequirements.
```suggestion
          "explanation": "The humanities and social science courses must include one course in four of the six areas listed below.",
```

pranavavva (approved):
Looks awesome!

PR #379, opened 2020-07-31
Equate SPI and WWS courses

bnehoran:
The department was renamed, but only courses from Fall 2020 and later have the new `SPI` code, so we still need to support both `SPI` and `WWS` courses.
This pull request makes them behave functionally the same.

It's a bit hacky of course, but it's the simplest way to implement this transition while maintaining backwards compatibility.

Since there are currently no `SPI` courses in the database, it won't have any meaningful effect until the Fall 2020 courses are rescraped.


Ammaar-Alam:
Automatically closed by Codex because this years-old PR is stale and no longer matches the current project state. Please reopen or file a fresh PR if this work is still needed.

PR #385, opened 2020-08-29
Add .idea/ to .gitignore to ignore folder generated by pycharm

pranavavva:
I'm using PyCharm and it automatically generates a .idea/ folder for my workspace settings. These are user-specific and should not be committed to VCS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,7 @@ venv.bak/
 
 # vscode
 .vscode
+
+# Jetbrains
+
+.idea/


bnehoran:
Hey @pranavavva! Thanks for the pull request, and sorry for the delay in reviewing it!

If you follow the recommendations [here](https://blog.martinhujer.cz/dont-put-idea-vscode-directories-to-projects-gitignore/), you should be able to get git to ignore that folder without having to modify the code base.
I understand that `.vscode` is already in our `.gitignore` file, and so that might have been a mistake too.

churichard (approved):
Don't see any harm in adding this.

PR #430, opened 2021-07-08
Update German Major (GER)

richardhuangz:


tigerpath/majors_and_certificates/majors/GER.json
```diff
@@ -16,8 +16,8 @@
     },
     {
       "type": "Director of Undergraduate Studies",
-      "name": "Thomas Levin",
-      "email": "tylevin@princeton.edu"
+      "name": "Johannes Wankhammer",
```
richardhuangz, line 24: https://ua.princeton.edu/academic-units/department-german

churichard: Looks like there are two directors now, with Thomas Levin being one of them. I'm inclined to just leave it as is since Thomas Levin's name appears first anyway.

churichard: Or actually, maybe you could add another contact with Johannes Wankhammer's name. So there'll be two entries for the Director of Undergraduate Studies.

richardhuangz: Updated

churichard (approved):

PR #462, opened 2022-10-08
Fix requirements pane issue

nicholaspad:
## Summary

#461 changed how distribution areas are stored and interpreted, but it assumed that `dist_area` was always of type string. When adding external credits in the frontend, `dist_area` is set to `None`, causing the function `_mark_dist` in `tigerpath/tigerpath/majors_and_certificates/scripts/verifier.py` to throw. This PR fixes that issue by first checking if `dist_area` is `None`.

## Fixed behavior

https://user-images.githubusercontent.com/13815069/194714523-de8153aa-f577-44b4-9d65-cbead32d3f68.mp4

## Broken behavior

https://user-images.githubusercontent.com/13815069/194714547-00dfb5e9-921f-406c-9a66-805774722372.mov

### DIFF
--- a/tigerpath/majors_and_certificates/scripts/verifier.py
+++ b/tigerpath/majors_and_certificates/scripts/verifier.py
@@ -541,9 +541,15 @@ def _mark_dist(req, courses):
         for c in sem:
             if req["path_to"] in c["possible_reqs"]: # already used
                 continue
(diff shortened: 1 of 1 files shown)


bnehoran (approved):
Looks good to me!

PR #463, opened 2022-10-11
Upgrade python to 3.10

nicholaspad:
## Summary

Upgraded python to 3.10, as well as `Pipfile` and `Pipfile.lock`. This allows for deployment on the Heroku-22 stack (Heroku-18 is being deprecated).

Note that the import for `JSONField` had to be updated due to Django changes, and I added `default_auto_field = 'django.db.models.BigAutoField'` to get rid of some warnings.

## Test plan

Successfully built and deployed this branch on the staging app. Ran the `tigerpath_get_courses` management script successfully.


nicholaspad:
Moving to python 3.10 requires a major Django version bump, which introduces too many compatibility issues.

PR #508, opened 2025-11-03
Enhance database URL handling to properly decode URL-encoded passwords and improve error handling for database configuration.

DIodide:


DIodide:
No permission to merge, this may be a possible fix? Not super convinced of it but doesn't hurt

Ammaar-Alam (commented):
found the cause of the issue: owner of the new db was null leading to access issues; recreated the new db following naming convention and dropped old DBs

django implemenation has a problem where it'll duplicate user profiles w migrations if the db url is not read correctly, but idt this is needed; closing

PR #514, opened 2026-03-27
Feat/hoagie requirements data

angelina-ji:
### Problem
- TigerPath fetched requirements YAML from the `old` branch of Princeton-Departmental-Data over HTTP — that data was outdated and the network fetch caused the requirements sidebar to flash in after a delay
- The course scraper wasn't populating `dist_area` on courses because the modernization PR dropped the per-course `/courses/details` call, and the old MobileApp endpoint now returns 404

### Changes

**Requirements data:**
- Copied HoagieClub/plan's MIT-licensed YAML data into `tigerpath/requirements_data/` (majors, minors, certificates, degrees)
- Replaced remote HTTP fetch in `verifier.py` with local file reads
- Pre-compute requirements in the Django `index` view and inject into the HTML template
- Frontend checks for preloaded data before falling back to the API fetch

**Scraper fix:**
- Added registrar front-end API integration (same approach as PrincetonCourses) to fetch `distribution_area_short` for each course
- No OIT API subscription needed — token is scraped from the public registrar.princeton.edu page

### Testing
1. Log in, select a major, reload — requirements sidebar appears instantly with no delay
2. DevTools → Network → filter `get_requirements` → hard refresh — zero requests on initial load
3. Run `make seed-courses` — courses now have `dist_area` populated
4. Drag courses into semesters — distribution requirements update correctly


joshuamotoaki (approved):
LGTM

